PR #10, opened 2015-02-08
Passing custom args to HandleFunc

dwilkie:
I have the following function which takes a transaction_id and msisdn and tries to create a Credit-ControlRequest:

```go
func Charge(transaction_id string, msisdn string, server_address string) (session_id string, result_code string) {

  dict.Default.Load(bytes.NewReader(dict.CreditControlXML))

  // ALL incoming messages are handled here.
  diam.HandleFunc("CEA", OnCEA) // I want to pass the transaction_id and msisdn through here
  diam.HandleFunc("CCA", OnCCA)
  diam.HandleFunc("ALL", OnMSG) // Catch-all.

  var (
    c   diam.Conn
    err error
  )
  c, err = diam.Dial(server_address, nil, nil)

  if err != nil {
    log.Fatal(err)
  }

  go NewClient(c)

  // Wait until the server kicks us out.

  <-c.(diam.CloseNotifier).CloseNotify()
  log.Println("Server disconnected.")
}

// I want this function to accept the transaction_id and msisdn as well so I can add it to the CCR AVP
func OnCEA(c diam.Conn, m *diam.Message) {
  rc, err := m.FindAVP(avp.ResultCode)
  if err != nil {
    log.Fatal(err)
  }
  if v, _ := rc.Data.(format.Unsigned32); v != diam.Success {
    log.Fatal("Unexpected response:", rc)
  }

  // Craft a CCR message.
  r := diam.NewRequest(diam.CreditControl, 4, nil)

// Add AVP's containing transaction_id and msisdn
}

func OnCCA(c diam.Conn, m *diam.Message) {
  // Here I want to return the result_code and session_id extracted from the AVP's
}
```

I can't set `transaction_id` or `msisdn` to a global variable, since this will cause problems with concurrency. So the only solution I can come up with at the moment is to send both the `CER` and `CCR` in the same function and hardcoding for example `DestinationRealm` instead of using the response in the `CEA`.

I looked through the RFC but I couldn't see anything in the spec for passing variables through - e.g sending `transaction_id` and `msisdn` in the `CER` and having it returned in the `CEA`.

Am I missing something simple here?

### DIFF
--- a/examples/client/client.go
+++ b/examples/client/client.go
@@ -75,7 +75,7 @@ func NewClient(c diam.Conn) {
 	ip, _, _ := net.SplitHostPort(laddr.String())
 	m.NewAVP(avp.HostIPAddress, avp.Mbit, 0, format.Address(net.ParseIP(ip)))
 	m.NewAVP(avp.VendorId, avp.Mbit, 0, VendorId)
-	m.NewAVP(avp.ProductName, avp.Mbit, 0, ProductName)
+	m.NewAVP(avp.ProductName, 0, 0, ProductName)
 	m.NewAVP(avp.OriginStateId, avp.Mbit, 0, format.Unsigned32(rand.Uint32()))
 	m.NewAVP(avp.VendorSpecificApplicationId, avp.Mbit, 0, &diam.GroupedAVP{
 		AVP: []*diam.AVP{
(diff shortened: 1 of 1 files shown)


fiorix:
Thanks @dwilkie. I'm merging this but beware that I'm about to push a new version v3.0.0 that fix all warnings from `golint`, better documentation and much better test coverage. However, this upcoming change is backwards incompatible since all constants e.g. SessionId are now SessionID and so on. Also, the `avp/format` sub-package has been renamed to `datatype` to reflect the RFC. You will have to make small changes to your existing code or use `godep` to keep on v2.0.4.

fiorix:
Oh, one more. By default, the `dict.Default` dictionary now has both the Base Protocols and Credit Control in it. No longer needed to run them separate.

PR #11, opened 2015-02-12
V3

fiorix:
This change has much better documentation and test coverage, adds support for a test server analogous to `net/http/httptest`, and has a few bug fixes (most notably in Float64 and Unsigned64).

It fixes all warnings from `go vet` and `golint`, at the expense of being backwards incompatible. Constants like `SessionId` are now `SessionID` and so on. Also, the `avp/format` package has been renamed to `datatype` to reflect the RFC nomenclature.

The default dictionary parser `dict.Default` has both the Base Protocol (app 0) and Credit Control (app 4) applications, and the standalone Credit Control is gone.


fiorix:
See this @dwilkie.

dwilkie:
:+1:

PR #16, opened 2015-02-25
update server.go to handle CER and DWR correctly

jaroszan:
- Added OnDWR handler
- Set Origin-State-Id to be optional in CER, as specified in rfc6733, section 5.3.1

### DIFF
--- a/examples/server/server.go
+++ b/examples/server/server.go
@@ -45,6 +45,7 @@ func main() {
 	runtime.GOMAXPROCS(*t)
 	// Message handlers:
 	diam.HandleFunc("CER", OnCER)
+	diam.HandleFunc("DWR", OnDWR)
 	diam.HandleFunc("CCR", OnCCR)
 	diam.HandleFunc("ALL", OnMSG) // Catch-all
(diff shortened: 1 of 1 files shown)


fiorix:
I've added a couple of comments to this change, please take a look.

jaroszan:
Updated according to your suggestions, details in the commit.

fiorix:
You need a `sync.Mutex` to guard the `allowedPeers` map since concurrent access to that map will cause trouble. Go has optimizations for empty structs for those and other cases, thus make it `allowedPeers map[string]struct{}` to save some resources.

Also, you're never deleting from that map when the client disconnect. In order to do that you have to watch for disconnect events in a separate goroutine. You can do `<-c.(CloseNotifier).CloseNotify()` to block until a client disconnects, and when that happen you remove it from the map. :)

jaroszan:
As previously details in the commit :)

jaroszan:
Given that this example will soon be no longer valid I guess it should be closed without merging. Still it was interesting to me from educational perspective, thanks for the hints :)

Issue #20, opened 2015-04-21
Called-Station-Id AVP definition missing


taseppa:
Seems like only the rule 	<rule avp="Called-Station-Id" required="false" max="1"/> is added but definition is missing which causes the application to fail processing a message that contains such AVP.

I wonder what process has been used to create the dictionary (xml) files, maybe there could be some other missing AVP definitions?

fiorix:
I see the rule defined but this AVP isn't defined neither in RFC 6733 or RFC 4006. FWIW it doesn't have to be in the built-in dictionary, you can load your own on top of it. See the examples/client and server.

taseppa:
Yeah it's actually a 3GPP specific AVP and should be defined in tgpp_ro_rf.xml. By doing some blind testing all AVP rules I tested had a corresponding definition, so far this is the only AVP I have noticed that is missing a definition.

fiorix:
Ok, so we didn't use any particular script for creating the dictionaries. If you define a new AVP in the xml files in `diam/dict/testdata` and run the `diam/autogen.sh` script, it will be added to generated files. You can check by looking at `avp/codes.go` afterwards.

If you search the git log you'll find the TS.xxx spec for reference.

∀

jaroszan:
Called-Station-Id actually belongs to Network Access Server Application (https://tools.ietf.org/html/rfc7155 or previously 4005). 

Not all avps which have a rule in Ro-Rf dictionary are defined in that file, some refer to Credit-Control application and since credit-control dictionary is preloaded you don't have that problem. 

I think a better approach would be to define a separate dictionary for NAS and maybe preload it as well?

fiorix:
Yes, that's definitely better. If the `dict.Default` dictionary becomes too bloated it may degrade performance of applications that don't need Ro-Rf though. Gotta see how many more AVPs this new one brings in.

PR #21, opened 2015-07-13
Missing Origin-State-Id should not make DWR fail since it's optional

mspronk:
Ran into this issue when deploying the sample diameter server: The client I was dealing with was not providing an Origin-State-Id in the DWR commands, which seems to be according to spec.

Also as per RFC-3588, 5.5.1 + 8.16 (https://tools.ietf.org/html/rfc3588)

### DIFF
--- a/diam/sm/dwr.go
+++ b/diam/sm/dwr.go
@@ -15,8 +15,8 @@ import (
 
 // handleDWR handles Device-Watchdog-Request messages.
 //
-// If mandatory AVPs such as Origin-Host, Origin-Realm, or
-// Origin-State-Id are missing, we ignore the message.
+// If mandatory AVPs such as Origin-Host or Origin-Realm are missing,
+// we ignore the message.
 //
 // See RFC 6733 section 5.5 for details.
 func handleDWR(sm *StateMachine) diam.HandlerFunc {
--- a/diam/sm/sm_test.go
+++ b/diam/sm/sm_test.go
@@ -58,6 +58,9 @@ func TestStateMachine(t *testing.T) {
 	mux.HandleFunc("CEA", func(c diam.Conn, m *diam.Message) {
 		mc <- m
 	})
+	mux.HandleFunc("DWA", func(c diam.Conn, m *diam.Message) {
+		mc <- m
+	})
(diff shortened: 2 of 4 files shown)


fiorix:
Great. I may have missed that. It's been a lot of typing to get this whole codebase in good shape. :)

fiorix:
Thanks!

mspronk:
Cool! Yeah implementing a whole RFC ain't easy ;-)

Thank YOU for all the work you put into this!

mspronk:
Hey Alexandre,

Do you happen to know off the top of your head whether the go-diameter
project has issues with nested Grouped AVP types?
I'm working on implementing a server with the 3GPP extension, part of which
is already in the base package you created, however I'm running into some
issues with building a sample client that contains nexted Grouped AVPs.

Thanks,
Martijn

fiorix:
Nested / grouped AVPs work just fine, I use it all the time for parsing and building messages.
Check out the example: https://github.com/fiorix/go-diameter/blob/master/examples/grouped/grouped.go

mspronk:
Awesome, thanks.

So I don't need to worry about the comment on line 36 here
<https://github.com/fiorix/go-diameter/blob/master/diam/group.go>, that says
"// TODO: handle nested groups?"

fiorix:
You could remove it and send a pull request :D

mspronk:
I definitely will ;-) .... I've also added a new 3GPP specific AVP that was
not there yet (Called-Station-Id) which i will add.

However I am still looking into a problem related to Vbit vs Mbit which I
want to resolve before committing the changes.
Basically when I send a sample CCR with Vbit set (the way the real-world
request is from the client, for the ServiceInformation AVP) ,
...then it fails on the server and the error seems to indicate a bit level
issue:

"Failed to decode AVP: Could not find AVP 0" , or..
"Failed to decode AVP: Could not find AVP 2147483676"

So I'm trying to understand that issue and hoping to make more of a
contribution that way ;-)

Cheers,
Martijn

mspronk:
Ignore that last "bug" it turned out to be a developer-issue ;-)

jaroszan:
For Called-Station-Id please see this issue: https://github.com/fiorix/go-diameter/issues/20. It is NAS specific.

PR #26, opened 2015-08-20
Support for searching grouped avps, support for search to return multiple avps

taseppa:
Unmarshal is functionality is nice when you want to parse multiple AVPs from a message. But in case you want to for example get only one AVP that is deeply nested inside grouped AVPs then unmarshal is a bit clumsy. This commit adds the ability to search within grouped AVPs and also added new search function which can return multiple AVPs.


fiorix:
Great, thanks!

fiorix:
I've made some other minor changes to data locality (methods within the Message type) for better readability. No functional changes, please pull it before continuing any work.

fiorix:
Had to do some cleanups on your commit to fix warnings from golint: https://github.com/fiorix/go-diameter/commit/66fb092596791eb8fbf3fe41aa714a2d6e8bfb99

PR #28, opened 2015-10-18
nas dict + reference rfc 7155

jaroszan:
Initial version of NAS dictionary (7155) along with the compressed rfc, autogen.sh wasn't applied in this commit, but it seems that destination files are successfully updated when autogen script is run.

I haven't done any tests for this dictionary yet, though xml looks correct. I will most likely have time next week to look into it, there are still some issues to consider, eg. overlapping command definitions between base and nas and making proper tests for this case. 

Another question is whether all dictionaries should be preloaded?


jaroszan:
For now adding QoSFilterRule as datatype is pending

fiorix:
Can't you add it? Should be pretty straightforward.

jaroszan:
Sure, I will add it.

jaroszan:
Added QoSFilterRule, it is basically a modified copy of IPFilterRule, both are derived from OctetString and use the same charset. rfc 7155 also says that AVP QoS-Filter-Rule, which uses this datatype, is not recommended and recommends using AVPs from rfc 5777 for QoS. It seems also that this AVP is the only AVP that uses this datatype.

fiorix:
Good, at least we have it supported.

fiorix:
Thanks!

PR #31, opened 2015-11-21
Message.FindAVPsWithPath implementation

danbogos:
Hi @fiorix 
First of all,  many thanks for your awesome library, it really saved us quite some time not to reinvent the wheel.

This PR will add support to search AVPs on path (eg. in case of AVP group hierarchies).

Ta,
DanB


fiorix:
Good. I've made a minor fix to the indentation of the example, updated the contributors file and merged.

Thanks!

PR #33, opened 2015-12-01
Parser.FindAVPWithVendor, fixes #23

### LINKED ISSUE #23: Problem having vendor specific AVP with an existing AVP code
According to RFC:
"The AVP Code, combined with the Vendor-Id field, identifies the  attribute uniquely."

But currently I seem to have some parsing problems if you use a vendor specific AVP with a code that already exists. At least in avp/codes.go this looks a bit odd:

dictAVP, err := dictionary.FindAVP(application, a.Code)

Because the FindAVP doesn't take vendorId as parameter.

danbogos:
Hi @fiorix 

Please let me know what you think about the changes.

Thanks,
DanB


fiorix:
Easy enough to follow through... could you please make it pass `golint` before I review once more and ensure it won't break other things.

PR #34, opened 2015-12-02
With passing golint

danbogos:


danbogos:
Sorry, please ignore since I see that it does not pass travis. I am doing a strange development with symlinks on my end since you depend on your own packages.

Anyway, will send you another PR in few mins.

DanB

fiorix:
There's no reason to complicate, having GOPATH=$HOME is the simplest and most effective thing.

danbogos:
I understand ... just that we use glide with vendors and go-diameter is now
part of our dependencies, hence the need of symlinks in vendor towards
forked package.

DanB
Am 02.12.2015 14:45 schrieb "Alexandre Fiori" <notifications@github.com>:

Issue #39, opened 2015-12-09
Failed to decode AVPs with Vendor-Id


mspronk:
When trying to decode a message that has AVPs with vendor-ids, that are in the default.go dictionary, ReadMessage errors out with "`Failed to decode AVP: Could not find AVP XXX`".

Verified by adding a test case in diam/message_test.go for a message with a Service-Information AVP (873). This is defined in default.go, but missing the vendor-id. This seems to be the case for all AVPs defined in the variable `tgpprorfXML`.

Proposed solution is adding `vendor-id="10415"` to each of these AVPs in default.go.
(PS Interim solution that worked for me for now was adding a corrected version of the entire `tgpprorfXML` section to my custom dictionary).

Will create a pull request for this fix.

fiorix:
Well, the Vendor ID is there: https://github.com/fiorix/go-diameter/blob/master/diam/dict/default.go#L1906

Perhaps the dictionary parser is not using that value for all AVPs in that dictionary. I wouldn't want to have it on each AVP. Maybe this is related to the changes from @danbogos?

fiorix:
O well... looking again, the dictionary itself support multiple Vendor IDs, and each AVP can only support one. So I guess the only way to make it proper is by updating each AVP.

danbogos:
@fiorix,

In my implementation the vendorId as general for application was not considered but the vendor_id parameter in the avp definition. Looking in the documents I could find around, that was also proposed in some dictionary formats .... one vendor_id per AVP.

Cheers,
DanB

danbogos:
To talk facts, here is how we define them (and it works now both ways with go-diameter):
https://github.com/cgrates/cgrates/blob/master/data/diameter/dict/huawei/3gpp_vendor.xml#L5

DanB

fiorix:
Thanks Dan, that's the right thing to do. So for #40 the Ro/Rf dictionary is fixed now.

danbogos:
Great, thanks for confirming it since it is all new ground for me also.

DanB

jaroszan:
That's an interesting issue which I didn't encounter during tests. AVPs in my case were defined as follows and were encoded/decoded correctly:

    m.NewAVP(avp.ServiceInformation, avp.Mbit|avp.Vbit, 10415, &diam.GroupedAVP{
		AVP: []*diam.AVP{
			diam.NewAVP(avp.IMSInformation, avp.Mbit|avp.Vbit, 10415, &diam.GroupedAVP{
[...]

PR #40, opened 2015-12-09
Issue #39 - Failed to decode AVPs with Vendor-Id

mspronk:
Added vendor-id value to all 3GGP AVPs
Added test case for message with AVPs that use vendor-id


fiorix:
Note that some of the files are auto-generated, and `dict/default.go` is one of them. So I updated the actual dictionary file and re-generated the files.

PR #50, opened 2016-07-12
Looked into enabling timing out dialling.

micrypt:


fiorix:
Looks fine, would you do it for TLS too?

micrypt:
Sure thing. I'll sort that out tonight.

fiorix:
Any word here?

fiorix:
Maybe you @jaroszan ?

micrypt:
@fiorix I wasn't certain what API to use for this, as the `net` package doesn't expose a `DialSSLTimeout` but rather requires using a client within the `tls` package. Consequently, I was a bit tad as to what's preferable. Would a separate client be better, or a design that's inconsistent with the Go standard library?

fiorix:
It's all on `crypto/tls`. You can choose between `Client(netConn, tlsConfig)`, or `Dial`, or `DialWithDialer`.

micrypt:
@fiorix Perhaps I was unclear… The question is not the means to implement it, but how (design-wise). Would it be okay to just add a `DialSSLTimeout` on `sm.Client`?

fiorix:
Yeah, both `DialTimeout` and `DialTLSTimeout` on both packages `diam` and `diam/sm` I would think.

jaroszan:
@fiorix: I'll check the related issue, but it looks that micrypt is already on top of it :)

fiorix:
Alright, both #50 and #60 merged. Thanks @micrypt and @jaroszan!

micrypt:
Thanks, @jaroszan!

PR #51, opened 2016-07-18
OriginStateID issue and Marshal function 

gmd20:
When testing  freeDiameter's  relay agent , I found several issues of go-diameter:  

1.  "Application Id" 0xffffffff for relay agent should be supported. See rfc6733  "2.4.  Application Identifiers"

2.  OriginStateID should be consistent in all requests inclulding the DWR, otherwise the remote peer will suppose the local peer has reboot and try to reset the connection.

3.  And I have implement a Marshal function in reflect.go. I think it can simplify the "build request" code .   An example can be found here https://github.com/gmd20/go-diameter

### DIFF
--- a/diam/reflect.go
+++ b/diam/reflect.go
@@ -15,11 +15,11 @@ import (
 	"github.com/fiorix/go-diameter/diam/dict"
 )
 
-// tagOptions is the string following a comma in a struct field's "json"
+// tagOptions is the string following a comma in a struct field's "avp"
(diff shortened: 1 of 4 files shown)


fiorix:
Looks good, but can you fix the tests before I merge?

gmd20:
It seems that 1 line of the change was missing when I commit it using github web interface.
The compile errors  have been fixed now.

fiorix:
Cool, thanks!

PR #55, opened 2016-09-06
added support for application id 3

jaroszan:
I have added two lines in base dictionary:

    `<application id="3"> '
    '</application>`

Run autogen.sh script and it seems to work. At least client does not complain about acct-application-id 3 anymore. Unfortunately I do not have real cases to test it.


fiorix:
Test still failing?

diam/dict/util_test.go
```diff
@@ -18,9 +18,17 @@ func TestApps(t *testing.T) {
 	if apps[0].ID != 0 {
 		t.Fatalf("Unexpected app.ID. Want 0, have %d", apps[0].ID)
 	}
+	// Base accounting.
+	if apps[1].ID != 3 {
+		t.Fatalf("Unexpected app.ID. Want 0, have %d", apps[1].ID)
```
fiorix: fix: Want 3, have %d

fiorix:
A couple more comments to fix tests error messages.

jaroszan:
I wonder whether the following modification makes more sense (https://github.com/fiorix/go-diameter/blob/master/diam/sm/smparser/app.go), from:

`if id == 0xffffffff`

to:

`if id == 0xffffffff || id == 3`

fiorix:
Hmmm not sure either. Would have to revisit the RFC, can't recall this.

jaroszan:
This one is a subset of https://github.com/fiorix/go-diameter/pull/56 , closing

PR #56, opened 2016-09-08
fix for common apps handling

jaroszan:
Not supported apps are logged to console without specifying whether id is local or remote, however local can still be recognized by 'missing application' error. Connection is established when remote peer returns ids unknown locally, but common applications exist.


fiorix:
We shouldn't log to console from a library like this. Programmers that use this library will be confused by these messages and tracking them down, plus, they might use different logging mechanisms other than `log.*`.

If we have to log errors like this I'd like to have the same semantics of net/http, on the [Server](https://golang.org/pkg/net/http/#Server) struct for example, see the `ErrorLog`.

jaroszan:
I will remove console output from the library and look into failing tests. Logging could be part of separate pull request if I find time for this.

jaroszan:
It looks that parser does not distinguish between go-diameter being client or server. Recognizing go-diameter role seems to be crucial for correctly handling differences in supported applications and establishing a connection.

fiorix:
Sure, let's record the role somewhere and use it appropriately for parsing then. I'll be on and off for the next 10 days starting today, with limited internet access, but will try to review anything you put up.

jaroszan:
It seems that to be compliant with RFC the following additional changes are needed:
1. Add type property to existing dictionaries (which would require regeneration), for both credit-control and nas type would be 'auth', base accounting will probably land in the dictionary with type 'acct' 
2. Prepare a dataset with supported apps so it can be used to send all capabilities to peer in CEA to fulfill statement from RFC6733, p. 5.3:

> The sender of the Capabilities-Exchange-Answer
(CEA) SHOULD include all of its supported applications as a hint to
the receiver regarding all of its application capabilities.

I am using something like this for testing purposes:

```
type SupportedApp struct {
    id      uint32
    apptype string
    vendor  uint32
}
supportedApps := []*SupportedApp{}
loaded_apps := dict.Default.Apps()
for _, app := range loaded_apps {
    if app.ID == 0 {
        continue
    }
    newApp := new(SupportedApp)
    newApp.id = app.ID
    newApp.apptype = app.Type
    for _, vendor := range app.Vendor {
        newApp.vendor = vendor.ID
    }
    supportedApps = append(supportedApps, newApp)
}
```
For further use in (most likely) successCEA (https://github.com/fiorix/go-diameter/blob/master/diam/sm/cer.go) to prepare response if at least one common app is found (currently CEA is composed of every app id found in CER).

So far everything is in the early testing phase, as soon as I have something worth presenting I put it up here.

fiorix:
Cool, I really like where this is going. 💯 

Run `golint` on it too.

jaroszan:
Missed one package in tests (smparser), will take a look at it. There is still some space for improvement, but basic tests with remote peer look ok and sm tests pass as well.

jaroszan:
Ok, it looks I have made a hasty decision with substituting ErrNoCommonApplication and ErrNoCommonSecurity with ErrCapabilitiesExchange. What would be the preferrable way to handle such errors (in other words - what to fix - errors or tests for smparser)?

fiorix:
It's a can of worms, right? :)

I'd rather keep the correct errors and fix UTs.

jaroszan:
I have doubts regarding https://github.com/fiorix/go-diameter/blob/master/diam/sm/smparser/cea_test.go , it builds and parses two CERs whereas https://github.com/fiorix/go-diameter/blob/master/diam/sm/smparser/cer_test.go has no tests for CEAs. Looks like a typo to me, but asking just in case.

jaroszan:
I've rollbacked some of the changes and errors are almost the same as before :) , tests are all right now, however there are still 2 more issues I would like to address:
1. Add more tests to parser to handle CE parsing for both clients and servers (to test for both no common application and missing application errors)
2. Find a good place to keep isServer parameter so it can be managed from one place (preferrably for tests too)

jaroszan:
For issue number 1:
I looked through the tests and from my perspective they verify the functionality, but do not necessarily consider wider context, eg. CEA should not be checked for common application on the client side, client already advertised what he intended and received a reply from the server. Reply can be either success or no_common_application, application ids do not need parsing in this scenario.  CERs also do not need to be checked on the client side since he does not receive any. It seems that some of the tests can be removed, but I don't want to do it since I could be missing something and these tests have their purpose (programming is auxiliary in my occupation). Can we leave these tests as they are in this branch and come back to them later (in the separate issue)? The way I would do it is taking all the tests, leave only those that fit in the wider context and remove the rest. Identify which tests could be added so that wider context is more complete and add them.

For issue number 2:
I see 3 options:
1. Move role determination to library users, client and server examples initialize state machine with lines:
`mux := sm.New(cfg)`
or
`mux := sm.New(settings)`

Supplying extra argument would determine what is passed to handleCEA and handleCER functions. But user will have to manage something that he does not need to.
2. Supply handleCER and handleCEA with appropriate argument in https://github.com/fiorix/go-diameter/blob/master/diam/sm/client.go for client and in https://github.com/fiorix/go-diameter/blob/master/diam/sm/sm.go for server. This way the user of the library does not to manage roles and everything should function as it is now.

3. Leave current implementation, specific values may not look very nice in the code, but they do the job and user of the library does not have to care about them.

I can't see any other mechanism to manage it at the moment.

jaroszan:
P.S. I am aware that recalculating supported apps may look inefficient for each connect attempt, but in real life scenario capabilities exchange take a negligible amount of time in comparison to other message exchanges (also diameter server will accept mostly known and accepted peers which is ensured via different mechanisms and such sessions normally last for a long time).

diam/sm/smparser/app.go
```diff
@@ -19,9 +19,34 @@ type Application struct {
...
+type SupportedApp struct {
+	ID      uint32
+	AppType string
+	Vendor  uint32
+}
+
+// PrepareSupportedApps prepares a list of locally supported apps
+func PrepareSupportedApps() []*SupportedApp {
+	locallySupportedApps := []*SupportedApp{}
+	for _, app := range dict.Default.Apps() {
```
fiorix: i don't think we should assume that `dict.Default.Apps()` is used here. It should probably be a parameter to `PrepareSupportedApps`.

diam/sm/smparser/app.go
```diff
@@ -19,9 +19,34 @@ type Application struct {
...
+		}
+		locallySupportedApps = append(locallySupportedApps, addApp)
+	}
+	return locallySupportedApps
+}
+
 // Parse ensures all acct or auth applications in the CE
 // exist in this server's dictionary.
-func (app *Application) Parse(d *dict.Parser) (failedAVP *diam.AVP, err error) {
+func (app *Application) Parse(d *dict.Parser, isServer uint8) (failedAVP *diam.AVP, err error) {
```
fiorix: i don't like this uint8 param here, because all calls to `Parse` look confusing now... instead, we should have a custom type and enums (iota) defining the supported types, client or server

```
type Role uint8

const (
  Server Role = iota + 1
  Client
)
```

Something like that..

fiorix: Another option would be to keep this "role" internal and offer `ParseClient` and `ParseServer` functions.

diam/sm/smparser/cea.go
```diff
@@ -23,7 +23,7 @@ type CEA struct {
 }
 
 // Parse parses and validates the given message.
-func (cea *CEA) Parse(m *diam.Message) (err error) {
+func (cea *CEA) Parse(m *diam.Message, isServer uint8) (err error) {
```
fiorix: Same comments of `CER.Parse`.. either make a Role or offer `ParseClient` and `ParseServer` if possible.

diam/sm/smparser/cer.go
```diff
@@ -30,7 +30,7 @@ type CER struct {
 // is found, then it returns the failedAVP (with the application that
 // we don't support in our dictionary) and an error. Another cause
 // for error is the presence of Inband Security, we don't support that.
-func (cer *CER) Parse(m *diam.Message) (failedAVP *diam.AVP, err error) {
+func (cer *CER) Parse(m *diam.Message, isServer uint8) (failedAVP *diam.AVP, err error) {
```
fiorix: Same as previous comments about `isServer uint8`.

diam/avp/codes.go
```diff
@@ -59,28 +53,43 @@ const (
...
 	AoCSubscriptionInformation            = 2314
-	ApplicID                              = 1218
 	ApplicationPortIdentifer              = 3010
 	ApplicationProvidedCalledPartyAddress = 837
 	ApplicationServer                     = 836
 	ApplicationServerID                   = 2101
 	ApplicationServerInformation          = 850
 	ApplicationServiceProviderIdentity    = 532
 	ApplicationSessionID                  = 2103
+	ApplicID                              = 1218
```
jaroszan: This is actually from TGPP spec, name is indeed confusing, but it is there (TS 32.299 ). It looks modified since I have run autogen.sh script after adding application types to dictionaries. But it was there since TGPP specific avps have been added.

jaroszan:
I made an additional change of moving preparation of supported apps to sm.go, it was a minor effort and location seemed more natural. Plus sm now gets supported apps on start up.

fiorix:
Let me know when you think it's ready to review again.

jaroszan:
It is ready now  (I had some mess in the branches but it looks sorted out)

jaroszan:
Please wait with review, I still have to check something.

jaroszan:
Ok, the issue is not directly connected - https://github.com/fiorix/go-diameter/issues/57.

It also looks like validate function in https://github.com/fiorix/go-diameter/blob/master/diam/sm/client.go can benefit from supportedApps property added to sm in https://github.com/fiorix/go-diameter/pull/56/commits/d714415a92d8b42ed5470f1b9b76ccf432d78ed2 , however at the moment I would rather put it on TODO list (one of the reasons being that I am not sure if moving supported apps preparation to sm.go is a good idea from your perspective).

Apart from that I think this PR is ready for review (I will address further changes separately).

fiorix:
💯

Issue #42, opened 2016-02-01
Error "missing Origin-State-Id" whilst using diam/sm.


micrypt:
Attempting to use state machine in a similar fashion as the `client.go` example, results in an error, `missing Origin-State-Id`. There's no clear way to update the handshake CER with the AVP in question in the case that it is indeed missing. A snippet from the application is below: 

```
    smcli := &sm.Client{
        Dict:               dict.Default,
        Handler:            mux,
        MaxRetransmits:     3,
        RetransmitInterval: time.Second,
        EnableWatchdog:     true,
        WatchdogInterval:   5 * time.Second,
        AcctApplicationID: []*diam.AVP{diam.NewAVP(avp.AcctApplicationID, avp.Mbit, 0,
            AcctApplicationID),
        },
    }

    // Set message handlers.
    mux.HandleFunc("CCA", cl.ccaHandler)

    // Print error reports.
    go printErrors(mux.ErrorReports())

    connect := func() (diam.Conn, error) {
        c, err := smcli.Dial(serverAddress)
        cl.c = c
        return cl.c, err
    }
```

fiorix:
Sounds like the error you get is because the server you're connecting to does not respond with a CEA containing `Origin-State-Id`. Can you confirm that with a pcap?

micrypt:
@fiorix You're absolutely correct. It's sending back an error message (3099). 

```
   DIAMETER_INVALID_AVP_BITS          3009
      A request was received that included an AVP whose flag bits are
      set to an unrecognized value, or that is inconsistent with the
      AVP's definition.
```

Is there a way to override the CER within the state machine's `handshake()`?

micrypt:
@fiorix Curiously, the only way I've managed to get a valid CEA response thus far has been to completely remove the FirmwareRevision AVP.

https://github.com/fiorix/go-diameter/blob/master/diam/sm/client.go#L179

fiorix:
I see, the `Firmware-Revision` AVP is not mandatory but the state machine will always include it. We have to patch it and make this optional somehow. Not sure how I want this implemented, need to think about it. Can't do for e.g. if revision=0 does not include, but also don't want to add a boolean field to that struct. Ideas are welcome.

fiorix:
Closing this in favor of #44.

micrypt:
[...]

PR #58, opened 2016-09-27
Firmware-Revision AVP encoded incorrectly (AVP Mandatory flag must not be set)

pmcgleenon:
Firmware-Revision AVP is being encoded incorrectly.   It has the AVP M bit set which is incorrect.

I have a go-diameter client talking to a server, and see this in the CER message from the go-diameter client.

The server side is complaining about Firmware-Revision AVP  with this code : DIAMETER_INVALID_AVP_BITS          3009

@micrypt describes the same problem in #42 

The AVP definition is correct -

<avp name="Firmware-Revision" code="267" must="-" may="-" must-not="P,V,M" may-encrypt="-">

but all of the code creating the FirmwareRevision AVP is setting the M bit


fiorix:
Great, thanks!

Issue #45, opened 2016-03-07
Unsupported address family: 0x0008


jaroszan:
When I receive a message I get an error as in subj.

Ethernet layer in wireshark says:

```
Type: IPv4 (0x0800)
```

IP layer is decoded as:

```
Internet Protocol Version 4, Src: 172.20.21.7, Dst: 172.20.17.37
0100 .... = Version: 4
.... 0101 = Header Length: 20 bytes
Differentiated Services Field: 0x00 (DSCP: CS0, ECN: Not-ECT)
Total Length: 580
Identification: 0xa4a1 (42145)
Flags: 0x02 (Don't Fragment)
Fragment offset: 0
Time to live: 62
Protocol: TCP (6)
Header checksum: 0x5a51 [validation disabled]
Source: 172.20.21.7
Destination: 172.20.17.37
[Source GeoIP: Unknown]
[Destination GeoIP: Unknown]
```

fiorix:
That's because we don't support 0x0800:

https://github.com/fiorix/go-diameter/blob/master/diam/datatype/address.go#L17

fiorix:
For reference, the RFC part for this specific data type:

https://tools.ietf.org/html/rfc6733#section-4.3.1

And the list of IANA address families:

http://www.iana.org/assignments/address-family-numbers/address-family-numbers.xhtml

jaroszan:
Initially I have looked at the wrong place, it seems I have found the AVP which causes the error (http://www.etsi.org/deliver/etsi_ts/132200_132299/132299/12.09.00_60/ts_132299v120900p.pdf#page=157):

> 7.2.170 Recipient-SCCP-Address
> The Recipient-SCCP-Address AVP (AVP code 2010) is of type Address. It is the "SCCP called address" used by the messaging node when delivering the message. This is usually the address of the MSC or SGSN/Serving Node that is serving the UE when it delivers the message. It contains a Global Title, where Global Title represents an E.164 number,  and possibly a Point Code (ISPC). The AddressType discriminator in RFC 3588 [401] is set to value 8, E.164, and the  address information is UTF8 encoded. 

Can you please advise what should be the scope of changes to make it work?

jaroszan:
Should Address be derived from OctetString, eg:

    type Address OctetString

?

And a case added for 0x08?

fiorix:
[...]

PR #59, opened 2016-09-29
support multiple address types

jaroszan:
This is the first draft, however I would like to ask for your take on approach taken here before proceeding.
1. GenericAddress is added without decode function and decode related tests, this domain should probably belong to Address type since diameter has no GenericAddress datatype. 
2. Decode function in address.go is responsible for decoding all datatypes except for reserved ones (0 and 65535)
3. Currently we are strict in what we receive, but not in what we send. Should Type interface have Serialize function which returns error as well? (for example verifying whether user supplies proper ipv4 and ipv6 address, currently it looks that anything that is not ipv4 is treated as ipv6 in address.go). In generic_address.go I have added print statements to signal errors, but that's definitely not what should be there.

I have made a basic test to make sure that what is added by GenericAddress is put on the wire in the same way as in the real network traces, but that's all the testing done so far with remote peers.


jaroszan:
Can you please review?

fiorix:
Hey @jaroszan I will try to find some time later this week to review. Sorry for the late response.

fiorix:
Could use some help from @dwilkie, @gmd20, @danbogos, @micrypt...

gmd20:
The "E.164 " address is rarely used?
In this special use case, you can simply declare the AVP as Octetstring.  It will allow you to specify any byte[].  And the helper functions like encodeE164/decodeE164  can be added into address.go

Or you can modify Address type to support E.164, but don't introduce a new type GenericAddress.
And please check if these changes affect the marshal and unmarshal functions in reflect.go.

fiorix:
Thanks @gmd20. Can we come to an agreement with @jaroszan?
Sorry for not being responsive in a timely manner here. Too much going on.

jaroszan:
@gmd20, @fiorix:
Actually GenericAddress was introduced in the first draft version (29 of september) and later removed (2nd of october) in https://github.com/fiorix/go-diameter/pull/59/commits/1b5d10d859059986be7b0e5195282e8780832ecd

Everything is in the Address datatype, it checks first whether input can be converted to IPv4 or IPv6 and adds AddressTypes (0x01 or 0x02). If Address datatype can't convert input to IPv4 or IPv6 address it treats input as bytes and does not add anything.  

As far as E164 usage is concerned, it is used quite often, for example charging mobile subscribers for various services in telecom sector (or not charging depending on the specific resource being used, or charging with special rates for particular resource)

I don't think is it possibly to reliably detect whether user input is e164. The current implementation keeps backwards compatibility and makes it possible to add very exotic datatypes. I briefly glanced through possible datatype formats and it does not look likely that some other datatype can be mistreated as IPv4 and IPv6, however without extensive testing one can't really be 100% sure.

Since no new datatype is introduced I can't see what can go wrong in reflect.go, but if you have something in mind please let me know.

gmd20:
@jaroszan,
   As you said, current implement doesn't allow the user to distinguish the IPV6 and E164 address.  So I recommend that you reserve the first two bytes (address family in byte[]) for all types of address, and provide the following interfaces:

func NewAddress(ip net.IP) *Address
func ParserIP (ip string)  Address 
func ParserE164 (e164 string)  Address 

func (addr Address)  AddressFamily()  uint16
func (addr Address)  ToIP () net.IP 
func (addr Address)  ToE164 () string

@fiorix may have other ideas, let's wait for his comments.

jaroszan:
@gmd20:
The first thing that I find difficult to agree with is that suggested solution does not account for address types different than IPv4, IPv6 and E164. Address types are plentiful (http://www.iana.org/assignments/address-family-numbers/address-family-numbers.xhtml) and it seems to me that it is easier to accomodate all in one shot rather than adding one by one and reevaluating existing solution each time before extending list of supported address types. But may be I am misunderstanging something?

In order to keep backwards compatibility (and it is very important if I understood correctly @fiorix in https://github.com/fiorix/go-diameter/issues/45) we can't really specify address types, it is detected based on user supplied input. If it is changed than backwards compatibility is broken.

May be you can provide complete example how library user is supposed to deal with addresses of different types (similarly to tests defined in address_test.go), it will definitely reduce potential misunderstanding on my side.

fiorix:
I read through this patch and I actually like it. Willing to merge unless @gmd20 has some more input.

gmd20:
It's OK with me.

fiorix:
👍

PR #60, opened 2016-09-30
adding timeout for tls dialing

jaroszan:
In support of https://github.com/fiorix/go-diameter/pull/50 , I have tried client with 1 and 0 nanosecond as timeout. It seems to work on my test server (1 nanosecond connection time out, while setting 0 seems to work like no timeout is set). I took the liberty to try it since last update for https://github.com/fiorix/go-diameter/pull/50 was 4 days ago, it does not replace, but complements pr 50.


fiorix:
Cool, thanks!

Issue #61, opened 2016-10-01
examples/server and examples/client don't connect


dirkm:
I tried to run examples/server/server.go and make a test-request using examples/client/client.go.

the server seems to come up:

```
dirk:~/gopath/src/github.com/fiorix/go-diameter/examples/server$ go run server.go
2016/10/01 09:28:28 Starting diameter server on :3868
```
In  another shell I open the client, but I am unable to connect
```
dirk:~/gopath/src/github.com/fiorix/go-diameter/examples/client$ go run client.go 
2016/10/01 09:40:50 Use wireshark to see the messages, or try -hello
2016/10/01 09:40:50 no common application
2016/10/01 09:40:52 no common application
```

Is this expected? I must admit that these are my first steps in go.

The server responds with an error to a simple telnet localhost 3868 as expected. I am using go1.6 on ubuntu 16.4

jaroszan:
I suggest that you comment out the following line in the client.go:

> diam.NewAVP(avp.AcctApplicationID, avp.Mbit, 0, datatype.Unsigned32(4)),  // RFC 4006

This example is obsolete since Credit Control application has type "auth" and this line adds application id with type "acct". go-diameter does not support application with id = 4 and type = acct. 

There is still an issue with error message here, client is adding application that he does not support. However if you want to send out what is supported by the dictionary it will be fine.

jaroszan:
You may also want to replace the following line:

`<application id="999">`
with
`<application id="999" type="acct">`
to make sure that server and client are on the same page.

More strict dictionary verification was introduced a couple of days ago, hence examples aren't up to date.

jaroszan:
See here for changes that will get examples to work https://github.com/fiorix/go-diameter/pull/62

dirkm:
Your fix works for me. Thanks for the quick patch!

fiorix:
Yeah no more hacks, #62 was just merged in. Thanks guys.

PR #63, opened 2016-10-02
removed dependency on parser for supported apps verification

jaroszan:
This one has the following goals:
- fix error message to user when client is initialized with apps unsupported locally (an issue introduced in https://github.com/fiorix/go-diameter/pull/56 and mentioned in https://github.com/fiorix/go-diameter/issues/61) - can be fixed by other means but current one seems natural
- make use of sm.supportedApps which is used for supported apps verification on the server side

It has 2 potential downsides:
- sm and sm client can be initialized with different dictionaries (however it does not seem likely)
- fmt dependency is added (though it makes it possible to generate very descriptive message to library user )

### DIFF
--- a/diam/sm/client.go
+++ b/diam/sm/client.go
@@ -14,7 +14,6 @@ import (
 	"github.com/fiorix/go-diameter/diam/avp"
 	"github.com/fiorix/go-diameter/diam/datatype"
 	"github.com/fiorix/go-diameter/diam/dict"
-	//"github.com/fiorix/go-diameter/diam/sm/smparser"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


jaroszan:
Ok, will have to fix some smparser tests.

jaroszan:
TestClient_Dial_MissingApplication is removed since it looks like it doesn't account for that Auth-Application-Id and Acct-Application-Id are optional. 

Can you please review this one?

fiorix:
👍 

Let me know if you're done or if there's more to this one.

jaroszan:
Looks done now. There is an extra fix for TestCEA not connected to this PR (it didn't make sense to open separate PR)

fiorix:
Excellent, thanks!

Issue #64, opened 2016-10-25
CCR - Event-Timestamp


DefunctExodus:
For some reason I am having trouble pulling this value out using the `avp` struct tags, I can clearly see it in my payload and I can see that it is being correctly decoded in `util.go`, But I just get an empty time back into my struct.

However it works if I do a `FindAVP(avp.EventTimestamp, 0)`

I am sure I am just missing something..but I am lost as to what I am messing up. any suggestions?

fiorix:
Can you paste the struct or at least the field? I was looking at the reflect tests and realised we don't actually have a test for `time.Time` although it should be supported. Perhaps you can try `*AVP`?

DefunctExodus:
sure so this is my struct I am trying to use (or a part of it anyway). I will give `*AVP` a shot to see what happens.
```
type ccr struct {
	EventTimestamp                time.Time      `avp:"Event-Timestamp"`
	SessionID                     string             `avp:"Session-Id"`
	OriginHost                    string             `avp:"Origin-Host"`
	OriginRealm                   string             `avp:"Origin-Realm"`
	DestinationRealm              string             `avp:"Destination-Realm"`
	AuthApplicationID             int32              `avp:"Auth-Application-Id"`
	ServiceContextID              string             `avp:"Service-Context-Id"`
	CCRequestType                 int32              `avp:"CC-Request-Type"`
	CCRequestNumber               int32              `avp:"CC-Request-Number"`
	DestinationHost               string             `avp:"Destination-Host"`
}
```

fiorix:
That's what I suspected. So, try first changing from `time.Time` to `datatype.Time` and you should get the right value. I believe we don't have the right conversion from AVP to `time.Time` in `reflect.go`. You could probably add that, and a unit test.

DefunctExodus:
actually I had tried `datatype.Time` and it gives me the same result

DefunctExodus:
```
type ccr struct {
	EventTimestamp                datatype.Time      `avp:"Event-Timestamp"`
	SessionID                     string             `avp:"Session-Id"`
	OriginHost                    string             `avp:"Origin-Host"`
	OriginRealm                   string             `avp:"Origin-Realm"`
	DestinationRealm              string             `avp:"Destination-Realm"`
	AuthApplicationID             int32              `avp:"Auth-Application-Id"`
	ServiceContextID              string             `avp:"Service-Context-Id"`
[...]

PR #66, opened 2016-12-08
fix-time-reflection

jaroszan:


jaroszan:
It seems that nothing was broken and datatype.Time is correctly unmarshalled. 

There is still an issue with time zones, I will look into it.

Should address https://github.com/fiorix/go-diameter/issues/64

fiorix:
That's cool. Should be easy to fix the timezone thing, and also (very important) to support not only `datatype.Time` but also `time.Time` for ease of use of the unmarshaller.

DefunctExodus:
This is really great.

jaroszan:
So far datatype.Time works, I will add tests to check if time.Time is properly unmarshalled and look into it as well.

fiorix:
👍

jaroszan:
Both time.Time and time.Datatype look ok now with tests. Will look into timezones next.

fiorix:
Amazing. I like the diff so far.

jaroszan:
Thanks, reflecting was quite tricky, I didn't do it before.

Should be good for review now. Time for both types is converted to UTC before checking against expected time.

fiorix:
Reflection is a bit tricky to wrap your head around, but I'm happy that you did it yourself and also learned from it.

fiorix:
Should we have a test that is not in UTC time, to ensure we can actually parse it correctly?

jaroszan:
Do you mean using `time.LoadLocation` ?

If I understand correctly we are receiving time and formatting it as we please (and some baseline is needed since we run tests in different timezones and travis uses UTC). If you have something specific in mind let me know and I'll look into it.

fiorix:
What I'm saying is to have a test that has the timezone different than UTC, preferably some weird timezone such as those with half hour offset, and ensure we parse it correctly to the equivalent UTC time. For instance we could have a test to parse the Bangaluru (India) timezone, which is UTC+5:30.

jaroszan:
I am confused, please correct me if I am wrong:

1. According to RFC 6733, time datatype uses NTP format and is tied to UTC
2. Hence time that we get from the wire is always UTC, it's not possible to get NTP time with an offset
3. Zone/offset matters when we examine time locally so it's only a question of formatting UTC time received from the network

fiorix:
Nevermind. :)

PR #69, opened 2017-01-18
Add more error details for CEA failure

aholappa:
_Result-Code_ was not enough to identify some error cases. In one case, the code is a generic "DIAMETER_UNABLE_TO_COMPLY" and the specific error is in the _Error-Message_ AVP:

```
	Result-Code {Code:268,Flags:0x40,Length:12,VendorId:0,Value:Unsigned32{5012}}
	Error-Message {Code:281,Flags:0x0,Length:40,VendorId:0,Value:UTF8String{Origin-Host already connected},Padding:3}
```

This PR:
- adds _Failed-AVP_ and _Error-Message_ to CEA struct
- exposes CEA in ErrFailedResultCode

### DIFF
--- a/diam/sm/smparser/cea.go
+++ b/diam/sm/smparser/cea.go
@@ -29,13 +29,12 @@ type CEA struct {
 // ErrFailedResultCode is returned by Dial or DialTLS when the handshake
 // answer (CEA) contains a Result-Code AVP that is not success (2001).
 type ErrFailedResultCode struct {
-	Code uint32
-	Cea  *CEA
+	*CEA
 }
 
 // Error implements the error interface.
-func (e *ErrFailedResultCode) Error() string {
-	return fmt.Sprintf("failed Result-Code AVP: %d", e.Code)
+func (e ErrFailedResultCode) Error() string {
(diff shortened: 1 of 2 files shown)


diam/sm/smparser/cea.go
```diff
@@ -21,13 +21,16 @@ type CEA struct {
...
+	FailedAVP                   []*diam.AVP               `avp:"Failed-AVP"`
+	ErrorMessage                string                    `avp:"Error-Message"`
 	appID                       []uint32                  // List of supported application IDs.
 }
 
 // ErrFailedResultCode is returned by Dial or DialTLS when the handshake
 // answer (CEA) contains a Result-Code AVP that is not success (2001).
 type ErrFailedResultCode struct {
 	Code uint32
+	Cea  *CEA
```
fiorix: We should call this field CEA (not Cea) and remove the Code field due to duplication. I wouldn't mind a `Code()` method in this struct as a shortcut to return `CEA.ResultCode`.

aholappa:
- Changed CEA to be an anonymous struct field. This way the fields can be accessed directly via the error struct (e.ResultCode, e.OriginHost etc.)
- Added test case for FailedAVP

fiorix:
👍

PR #71, opened 2017-02-04
Nil pointer fix for sm/cer.go

aholappa:
I'm making some client-server test cases and got a panic on `github.com/fiorix/go-diameter/diam/sm/cer.go:80`. I'm missing the _Origin-host_ AVP and that causes an error of type _errors.errorString_ which wasn't handled by the switch case.


fiorix:
👍

PR #73, opened 2017-02-13
Allow clients to provide a pre-defined ip address.

VDVsx:
In some cases the requests might be fired from different machines via a gateway
or a NAT, local IP might not match the rules defined in the diameter server side.

### DIFF
--- a/diam/sm/sm.go
+++ b/diam/sm/sm.go
@@ -41,10 +41,11 @@ func PrepareSupportedApps(d *dict.Parser) []*SupportedApp {
 // Settings used to configure the state machine with AVPs to be added
 // to CER on clients or CEA on servers.
 type Settings struct {
-	OriginHost  datatype.DiameterIdentity
-	OriginRealm datatype.DiameterIdentity
-	VendorID    datatype.Unsigned32
-	ProductName datatype.UTF8String
+	OriginHost   datatype.DiameterIdentity
+	OriginRealm  datatype.DiameterIdentity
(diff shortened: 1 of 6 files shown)


VDVsx:
Not that familiar with this code yet, if some changes are needed somewhere else please let me know. Thanks.

fiorix:
The `Settings` struct is used for both CER and CEA. We should honor the custom host IP address on both, not only CER.

VDVsx:
@fiorix Thanks, added now to CEA as well, added also some simple tests covering the added setting.

diam/sm/sm.go
```diff
@@ -41,10 +41,11 @@ func PrepareSupportedApps(d *dict.Parser) []*SupportedApp {
 // Settings used to configure the state machine with AVPs to be added
 // to CER on clients or CEA on servers.
 type Settings struct {
-	OriginHost  datatype.DiameterIdentity
-	OriginRealm datatype.DiameterIdentity
-	VendorID    datatype.Unsigned32
-	ProductName datatype.UTF8String
+	OriginHost   datatype.DiameterIdentity
+	OriginRealm  datatype.DiameterIdentity
+	HostIPAdress datatype.Address
```
fiorix: Just an aesthetic change that might be important: let's move this field to the bottom of the struct and add comments explaining that it is optional and the use case for it.

VDVsx: @fiorix done :)

fiorix:
👍

PR #74, opened 2017-03-23
Allow a custom DWA handler if watchdog is not enabled

aholappa:


fiorix:
LGTM but maybe we should document this somewhere along with the code change?

fiorix:
@aholappa ping

aholappa:
Added quick documentation to code comments.

PR #76, opened 2017-11-01
Fix for unit tests on Go 1.9

emakeev:
Fixes issue with AVP Enum declared as uint8: https://github.com/fiorix/go-diameter/blob/master/diam/dict/skel.go#L75 and
some TGPP AVPs using enum values > 255: https://github.com/fiorix/go-diameter/blob/master/diam/dict/default.go#L3515-L3523

Updated test server cert & key: https://github.com/fiorix/go-diameter/blob/master/diam/diamtest/server.go#L118-L127

Tested on Mac OS 10.12.6 with Go 1.9


fiorix:
Could you also please add newer go versions to .travis.yml to have this in CI, in the same diff?

diam/dict/default.go
```diff
@@ -13,11 +13,28 @@ import "bytes"
...
-	Default.Load(bytes.NewReader([]byte(baseXML)))
-	Default.Load(bytes.NewReader([]byte(creditcontrolXML)))
-	Default.Load(bytes.NewReader([]byte(networkaccessserverXML)))
-	Default.Load(bytes.NewReader([]byte(tgpprorfXML)))
+	var err error
+
+	Default, err = NewParser()
+	if err != nil {
+		panic(err)
+	}
```
fiorix: Don't really think this is needed but if you want to go this route let's please loop over an array

diam/dict/util.go
```diff
@@ -61,17 +61,18 @@ retry:
 	case string:
 		avp, ok = p.avpname[nameIdx{appid, code.(string), vendorID}]
 		if !ok && appid == 0 {
-			err = fmt.Errorf("Could not find AVP %s", code.(string))
+			err = fmt.Errorf("Could not find AVP string %s for VendorID: %d", code.(string), vendorID)
```
fiorix: Maybe using %q instead of %s makes it more readable

fiorix:
Thanks !

PR #77, opened 2017-11-15
Added DialConn for passing any net.Conn to client.

rakshasa:
Allow passing any net.Conn to go-diameter in order to support other protocols such as sctp.

### DIFF
--- a/diam/client.go
+++ b/diam/client.go
@@ -100,3 +100,15 @@ func dialTLS(srv *Server, certFile, keyFile string, timeout time.Duration) (Conn
 	go c.serve()
 	return c.writer, nil
 }
+
+// DialConn is the same as Dial, but using an already open net.Conn.
+func DialConn(rw net.Conn, addr string, handler Handler, dp *dict.Parser) (Conn, error) {
+	srv := &Server{Addr: addr, Handler: handler, Dict: dp}
+
+	c, err := srv.newConn(rw)
+	if err != nil {
+		return nil, err
+	}
+	go c.serve()
+	return c.writer, nil
+}
--- a/diam/client.go
+++ b/diam/client.go
@@ -101,8 +101,8 @@ func dialTLS(srv *Server, certFile, keyFile string, timeout time.Duration) (Conn
 	return c.writer, nil
 }
 
-// DialConn is the same as Dial, but using an already open net.Conn.
-func DialConn(rw net.Conn, addr string, handler Handler, dp *dict.Parser) (Conn, error) {
+// NewConn is the same as Dial, but using an already open net.Conn.
+func NewConn(rw net.Conn, addr string, handler Handler, dp *dict.Parser) (Conn, error) {
 	srv := &Server{Addr: addr, Handler: handler, Dict: dp}
 
 	c, err := srv.newConn(rw)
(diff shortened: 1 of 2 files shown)


diam/client.go
```diff
@@ -100,3 +100,15 @@ func dialTLS(srv *Server, certFile, keyFile string, timeout time.Duration) (Conn
 	go c.serve()
 	return c.writer, nil
 }
+
+// DialConn is the same as Dial, but using an already open net.Conn.
+func DialConn(rw net.Conn, addr string, handler Handler, dp *dict.Parser) (Conn, error) {
```
fiorix: Can we call it NewConn instead, just to convey that we're creating a new connection from an existing net.Conn.

Also, we don't need the `addr` param, you can get it from `rw.RemoteAddr().String()`.

diam/sm/client.go
```diff
@@ -86,6 +86,13 @@ func (cli *Client) DialTLSTimeout(addr, certFile, keyFile string, timeout time.D
 	})
 }
 
+// DialConn is like Dial, but using an already open net.Conn.
+func (cli *Client) DialConn(rw net.Conn, addr string) (diam.Conn, error) {
```
fiorix: Ditto

diam/client.go
rakshasa: I initially used ServeConn before changing it for consistency with the other functions.

It doesn't actually make a new connection, rather it passes it to the client so it can serve it, so might that be more descriptive?

fiorix: Doesn't Dial either. Naming is hard!

rakshasa: Serve is the one that seems most logical to me, as new feels as if you are getting something from the target, rather than making it do something.

fiorix: Looking at the three options we discussed so far:

```
NewConn(net.Conn, Handler, *dict.Parser) (Conn, error)

DialConn(net.Conn, Handler, *dict.Parser) (Conn, error)

ServeConn(net.Conn, Handler, *dict.Parser) (Conn, error)
```

I still think `NewConn` makes more sense because it's creating a new `diam.Conn` after all. But I'm ok with `ServeConn` too.

What about removing the `addr` parameter?

rakshasa: The sctp library I'm currently using and might end up in the official go 'x' repository, it doesn't behave as simply as regular tcp/udp protocols when it comes to getting local and/or remote addresses.

When connected, with an fd, the sctp protocol seems to be able to have multiple sockaddr on either side, and can add/remove them without closing the connection.

So a RemoteAddr() call could give you an empty string, the normal address or a '/' separated list.

rakshasa: Which means that when the sctp thing is ready it will likely be a separate sub-package in go-diameter with a dedicated dial function.

Since we don't know how things will end up, I'd say exposing the addr is the better route until better api's are ready.

fiorix: Makes sense to keep the address then. I'll merge as soon as you rename the Dial functions.

Where is the current sctp code? I'd like to take a look :)

rakshasa: The code is currently in a private repository, so won't be available for now.

rakshasa:
Forgot to poke, fixed the names.

fiorix:
👍

PR #79, opened 2018-03-15
Add 3GPP S6a commands Notify and Purge-UE

bearmini:
- add commands Notify (NOR/NOA) and Purge-UE (PUR/PUA)
- add their AVPs
- modify authgen.sh to get work on mac as same as on linux


fiorix:
ping @bearmini

fiorix:
@emakeev could you please take a quick look at these diffs to make sure it won't break your last diff? otherwise @bearmini will have to rebase to master.

fiorix:
Merging. Thanks!

PR #85, opened 2018-07-13
close correct connection in cleanupConn

dirkm:
I think this patch fixes a minor issue when releasing a connection. With the patch, we never call close a nil-object.

Note: this has been found from manual inspection, we did not reproduce the failure.


fiorix:
Thanks!

PR #87, opened 2018-07-18
Support multiple Host-IP-Address APVs in CER and CEA

bearmini:
CER and CEA definitions require Host-IP-Address AVP as following in RFC 6733 Diameter Base Protocol:
`1* { Host-IP-Address }`
This means Host-IP-Address AVP should appear at least once, and it is allowed to appear multiple times.

In the real world HLR usage, we need multiple Host-IP-Address APVs for redundancy.


fiorix:
👍 LGTM.

cc @emakeev

PR #90, opened 2018-07-25
support binding local address

bearmini:
This PR enables to use `sctp.DialSCTP()` with laddr (local address) to bind a particular IP address and a port number on client side.
Some DRAs requires their clients to use a specified source IP address and port number.


fiorix:
Going to merge this, but want to notify @emakeev that it seems slightly different from another diff I've seen recently adding the same functionality in vendored code.

PR #92, opened 2018-07-30
Fix missing Mbit when must=M,V in AVP spec

higebu:


higebu:
This PR fixes AVPs like this:
https://github.com/fiorix/go-diameter/blob/master/diam/dict/default.go#L5026

fiorix:
👍

PR #97, opened 2018-09-18
Multistreaming SCTP support & additional 3GPP protocol definitions

emakeev:
Adding multistreaming SCTP support (current SCTP implementation is using default (0) stream only) and some additional 3GPP AVPs/commands.


fiorix:
Couple of nits but no show stoppers. Merging.

PR #109, opened 2019-07-16
Make ResultCode optional when creating an Answer from Message

TeoV:
According to RFC 4006 Section 3.2 ( https://tools.ietf.org/html/rfc4006#section-3.2 ) the first field should be Session_Id not ResultCode.
With this PR the ResultCode AVP is now optional.


fiorix:
I squashed the two commits into one. Thank you !

PR #120, opened 2019-12-27
Extending SWx support; Go 1.13.5 TLS unit test fixes; CE multi vendordor App IDs

emakeev:
... also IPv6 data type & minor fixes


diam/diamtest/server.go
```diff
@@ -70,6 +70,7 @@ func newLocalListener(network string) net.Listener {
 	}
 	l, err := diam.MultistreamListen(network, "127.0.0.1:0")
 	if err != nil {
+		fmt.Printf("diamtest: failed initial listen on network %s: %v", network, err)
```
fiorix: does this belong here?

PR #131, opened 2020-04-13
Add support for context for diam.Message

higebu:
Context per diam.Message is useful for creating middlewares for go-diameter.


emakeev:
Could you please explain what adding context to the message accomplishes? Maybe unit tests / examples.

higebu:
@emakeev Thank you for your comment. I'll add unit test soon.

higebu:
@emakeev I added unit tests and example.

diam/message.go
```diff
@@ -575,3 +577,17 @@ func indentTabs(n int) string {
...
 	return s
 }
+
+// Context returns the message's context. To change the context, use
+// SetContext.
+func (m *Message) Context() context.Context {
+	if m.ctx != nil {
+		return m.ctx
+	}
+	return context.Background()
```
emakeev, line 587: Could you change the logic to not to return context.Background() if the ctx is not set. It's a bit confusing (and also - is not reflected in the function documentation). The way, it's implemented now, one can call m.SetContext(nil), then m.Context() and get a different result (not previously set nil).
I understand that this will require an extra checks & set in your example use case, but it's probably a reasonable price to pay for clarity of the API:

```
func (m *Message) Context() context.Context {
	if m != nil {
		return m.ctx
	}
	return nil
}
```

or just

```
func (m *Message) Context() context.Context {
	return m.ctx
}
```

emakeev:
@higebu thank you for providing unit test & example. Could you please address one issue (please, see the inline comment)?

diam/message.go
emakeev: NM, I guess it corresponds to Go's HTTP ctx implementation.

PR #146, opened 2020-11-13
Customizable dictionary and list of supported applications for CEA

Yoone:
I need to specify the list of actual Application IDs / Vendor IDs my equipment actually supports. In order to do so, I had to override the existing `Default` dict, which isn't possible everywhere, namely in the default `StateMachine` creation (`sm.New` uses `dict.Default`).

On top of that, declaring a new `*dict.Parser` is cumbersome since all I want is to use the default one but with less interfaces. So I made all the default XML dicts public, and added a helper to initialize a dictionary using those.

With the changes in the PR, one can now change the default dict (and associated Application IDs / Vendor IDs) this way, reusing the dictionaries already declared in `go-diameter`:

```go
import (
    "github.com/fiorix/go-diameter/v4/diam/dict"
    "github.com/fiorix/go-diameter/v4/diam/dict/library"
    "github.com/fiorix/go-diameter/v4/diam/sm"
)

dictionary, err := dict.NewParserFromLibrary(
    library.Base,
    library.CreditControl,
    library.TgppRoRf,
)
if err != nil { ... }

settings := &sm.Settings{
    // some settings
    Dictionary: dictionary,
}
stateMachine := sm.New(settings)
```

I did move all the XML (some 6k lines) to a new package because I believed it made more sense with the new structure, but I can revert that and make my changes in place if that's better.


emakeev (changes requested):
Dictionary .go files are generated from dictionaty definition XML files in https://github.com/fiorix/go-diameter/tree/master/diam/dict/testdata by this script: https://github.com/fiorix/go-diameter/blob/master/diam/autogen.sh

This PR does not take it into account and will break the ability to autogenerate dictionaries.

fiorix:
cannot be merged

PR #148, opened 2020-11-24
Support graceful shutdown

jefftt:
This PR takes the minimal set of logic necessary to support graceful shutdown via a `Shutdown()` method from the [std net/http server](https://golang.org/src/net/http/server.go?s=85127:85181#L2689)

Note: this does **not** implement the diameter [peer disconnect](https://tools.ietf.org/html/rfc3588#section-5.4)

### DIFF
--- a/diam/server.go
+++ b/diam/server.go
@@ -143,7 +143,7 @@ func (c *conn) notifyClientGone() {
 }
 
 // A ConnState represents the state of a client connection to a server.
-type ConnState int
+type ConnState int8
 
 const (
 	// StateNew represents a new connection that is expected to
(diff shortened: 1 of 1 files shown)


diam/server.go
```diff
@@ -137,6 +142,45 @@ func (c *conn) notifyClientGone() {
 	}
 }
 
+// A ConnState represents the state of a client connection to a server.
+type ConnState int
```
emakeev, line 146: why not int8 here?

diam/server.go
```diff
@@ -137,6 +142,45 @@ func (c *conn) notifyClientGone() {
...
+
+var stateName = map[ConnState]string{
+	StateNew:    "new",
+	StateActive: "active",
+	StateIdle:   "idle",
+	StateClosed: "closed",
+}
+
+func (c ConnState) String() string {
+	return stateName[c]
```
emakeev, line 181: name, ok := stateName[c]
if !ok {
       return "UNDEFINED"
}
return name

diam/server.go
jefftt: taken from https://github.com/golang/go/blob/master/src/net/http/server.go#L2823

looks like it can be an int8 afaict, can update

PR #149, opened 2020-12-02
Bug. Parsing 'omitempty' from another tag value

vildam:
If struct has several tags, the function `parseAvpTag` will get 'omitempty' value from json tag

Function will return ("Session-Id",true) for reflect.StructTag(`avp:"Session-Id" json:"session,omitempty")


emakeev (changes requested):
Build & unit tests are failing. Please see: https://travis-ci.org/github/fiorix/go-diameter/builds/747191732

Issue #172, opened 2022-10-27
Make certificates optional in ListenAndServeTLS to allow an external callback


AlexanderMescheryakov:
It seems that currently it's not possible to use a callback for fetching TLS certificates on the fly. It's useful for hot replacement of certificates without a need of an application restart. That might be possibly by providing a GetCertificate callback via the tls.Config input of the Server. But the issue is that the `ListenAndServeTLS(certFile, keyFile string)` unconditionally fills config.Certificates and fails if the certificate files are missing. This can be fixed by means of adding an option into ListenAndServeTLS to skip the lines:
https://github.com/fiorix/go-diameter/blob/master/diam/server.go#L697
```
config.Certificates = make([]tls.Certificate, 1)
config.Certificates[0], err = tls.LoadX509KeyPair(certFile, keyFile)
```
Or it might check if the config.GetCertificate is nil and load certificates only in that case.

PR #173, opened 2022-10-27
Do not load certificates if GetCertificate is provided

AlexanderMescheryakov:
Addresses #172


paulrozhkin:
I think this check should be like in the `net/http/server.go`. It contains:
```
configHasCert := len(config.Certificates) > 0 || config.GetCertificate != nil
if !configHasCert || certFile != "" || keyFile != "" {
	var err error
	config.Certificates = make([]tls.Certificate, 1)
	config.Certificates[0], err = tls.LoadX509KeyPair(certFile, keyFile)
	if err != nil {
		return err
	}
}
```

PR #176, opened 2023-06-07
Add GitHub Actions workflow for CI and some fixes

higebu:
I'd like to add a GitHub Actions workflow for CI, because [Travis CI link](http://travis-ci.org/fiorix/go-diameter) is broken.

- [Update go version to 1.20 in go.mod](https://github.com/fiorix/go-diameter/commit/20c7f6b0bc0a7fee6db93c141ac0aee53740d160)
- [Add github.com/ishidawataru/sctp to go.mod](https://github.com/fiorix/go-diameter/commit/6760c675b501923f1778cbfabef9dc28b8354a73)
- [Fix tests with tls](https://github.com/fiorix/go-diameter/commit/2213d4c7ce2863b8b5a8412d90b73a562da3724a)
- [Use go:build style build constraints](https://github.com/fiorix/go-diameter/commit/bd6f8ee36dd958ddf49de831d1cbcd9893167fef)
- [Use context pkg instead of golang.org/x/net/context](https://github.com/fiorix/go-diameter/commit/1b7c819c579ef5067b6856b06590e7c883ec9de6)
- [Add GitHub Actions workflow for CI](https://github.com/fiorix/go-diameter/commit/a9a6a3b746d0ea8062ce7820b9c868c519db8c88)
- [Fix panic getLocalAddresses() on sctp](https://github.com/fiorix/go-diameter/commit/66d104d0ad4b946685d64b5ee84f575eb51b91e4)

PR #178, opened 2023-07-03
Update README

higebu:
* Add Go Report Card, Test Status, Go Reference, and Latest version badges.
* Use go.dev for official Go links.

Related to: #176


fiorix:
Thank you!

PR #191, opened 2023-12-19
fix: properly parse avp struct tag

SX91:
Read `omitempty` from `avp` tag instead of the end of the tags string.


SX91:
@fiorix could you please review this so it could be merged or closed?

This allows user to add more than just avp tag to fields. Like `json`.

Issue #180, opened 2023-07-06
Missing Handshake Notifications in channel


SX91:
Hello!
I've used the approach similar to the [proposed one here](https://github.com/fiorix/go-diameter/issues/22#issuecomment-122586024), but sometimes the connected peer is not being registered.
Seems like it happens because the `StateMachine.hsNotifyc` is created as unbuffered channel, and writes to this channel are non-blocking (select + default), thus "skipping" some notifications if the receiver is "slow".

Any suggestions how to make notifications reliable?

AndrewK1273:
I have the same experience. When two clients connect at the same time, the handshake notification from one of them is missing

fiorix:
Happy to try and review a patch to resolve this

AndrewK1273:
Would it be a problem to make the hsNotifyc channel buffered? If there were no consumers, a few messages would remain in the channel, but that wouldn’t matter, would it?

PR #202, opened 2024-11-12
creating hsNotifyc as a buffered channel

AndrewK1273:
If two clients connect at the same time, an HS notification may be lost because writing to the channel is non-blocking and the channel is not buffered.

### DIFF
--- a/diam/sm/sm.go
+++ b/diam/sm/sm.go
@@ -96,7 +96,7 @@ func New(settings *Settings) *StateMachine {
 	sm := &StateMachine{
 		cfg:           settings,
 		mux:           diam.NewServeMux(),
-		hsNotifyc:     make(chan diam.Conn),
+		hsNotifyc:     make(chan diam.Conn, 10),
 		supportedApps: PrepareSupportedApps(dict.Default),
 	}
 	sm.mux.Handle("CER", handleCER(sm))


diam/sm/sm.go
```diff
@@ -96,7 +96,7 @@ func New(settings *Settings) *StateMachine {
 	sm := &StateMachine{
 		cfg:           settings,
 		mux:           diam.NewServeMux(),
-		hsNotifyc:     make(chan diam.Conn),
+		hsNotifyc:     make(chan diam.Conn, 10),
```
fiorix, line 99: Were you able to reproduce the problem and test with this change? Also I'm curious for what makes "10" the right size for the buffered channel here.

AndrewK1273:
It depends on how robust it should be.

When I test it with 10 clients connecting simultaneously, which is probably not very likely in the real world, notifications stop being lost when the buffer set to 3 (see attachment).

It also depends on the way notifications are processed on the server side. During the test, I print out originHost.

func (f *FakePcrf) HandshakeNotifyHandler(hn <-chan diam.Conn) {
	for c := range hn {
		if meta, ok := smpeer.FromContext(c.Context()); ok {
			log.Printf("new peer OriginHost: %s", string(meta.OriginHost.Serialize()))
		}
	}
}

The value of 10 seemed like a sufficient reserve even for a very crazy case.
[diam_hs_notif_tests.txt](https://github.com/user-attachments/files/17735488/diam_hs_notif_tests.txt)

SX91:
@fiorix with the current unbuffered channel notifications are getting dropped on the sender side if receiver is not reading them right away. #180 
Would it be better to add Options support to New providing the custom channel size?

PR #206, opened 2025-03-27
Br diam lib fiorix path rename

deb-2025:
diam lib fiorix path renamed to deb-2025


fiorix:
Changing the import paths would break my repo, won't be able to merge it.

PR #211, opened 2025-05-14
Adding Cx interface support 

rakeshgmtke:
Adding Cx interface support


rakeshgmtke:
Im adding CX interface related AVP and Autogen.sh updated

Issue #208, opened 2025-04-11
autogen.sh script update required handle \n


Rakeshgmtke:
autogen.sh script can insert a new line with echo above 8.30. 

It's updated in latest echo version.

/bin/echo --version
echo (GNU coreutils) 8.30
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Brian Fox and Chet Ramey.

**echo -e**  

https://unix.stackexchange.com/questions/189787/difference-between-echo-and-echo-e

**from echo manual pages**

       If -e is in effect, the following sequences are recognized:

       \\     backslash

       \a     alert (BEL)

       \b     backspace

       \c     produce no further output

       \e     escape

       \f     form feed

       \n     new line

       \r     carriage return

       \t     horizontal tab

       \v     vertical tab

       \0NNN  byte with octal value NNN (1 to 3 digits)

       \xHH   byte with hexadecimal value HH (1 to 2 digits)

       NOTE: your shell may have its own version of echo, which usually supersedes the version described here.  Please refer to your shell's documentation for details about the options
       it supports.

PR #212, opened 2025-05-15
update autogen.sh and Rx xml addition

rakeshgmtke:
1. Update Autogen.sh scipt to handle new line with echo
2. Rx xml added to and generated avp/codes/application file update with autogen.sh script


rakeshgmtke:
Above changes can fix issue **https://github.com/fiorix/go-diameter/issues/208**

fiorix:
This cannot be merged until you update the package path back to the original. All of your code is now using your own clone and this change would break this repo.

fiorix:
Thanks @rakeshgmtke for the effort on adding Rx and Cx dictionary support — that's valuable work!

Unfortunately this PR can't be merged as-is due to several issues:

1. **Module path changed to your fork** — all imports point to `github.com/rakeshgmtke/go-diameter` instead of `github.com/fiorix/go-diameter/v4`, which would break the module for all users
2. **Duplicate constants** — `MultimediaAuth` conflicts with existing `MultimediaAuthentication` (both = 303), and `FramedIpv6Prefix` conflicts with `FramedIPv6Prefix` (both = 97)
3. **`echo -e` portability** — the script uses `#!/bin/sh` but `echo -e` is a bashism; should use `printf`
4. **Large diff from reordering** — regenerated files reorder existing entries making review difficult

If you'd like to rework this, a fresh PR with the correct module path and the above issues addressed would be welcome. Thanks again for the contribution!

dunst0:
@fiorix Quick question since I'm just checking this PR here as I would like to use this library for some little project.

From reading the specs for SWx and Cx/Dx I would say the name for the command would be better MultimediaAuth than MultimediaAuthentication.

Cx/Dx interface description from 3GPP TS 29.229
```
< Multimedia-Auth-Request > ::= < Diameter Header: 303, REQ, PXY, 16777216 >
                                < Session-Id >
                                { Vendor-Specific-Application-Id }
                                { Auth-Session-State }
                                { Origin-Host }
                                { Origin-Realm }
                                { Destination-Realm
```

SWx interface description from 3GPP TS 29.273
```
< Multimedia-Auth-Request > ::= < Diameter Header: 303, REQ, PXY, 16777265 >
                                < Session-Id >
                                [ DRMP ]
                                { Vendor-Specific-Application-Id }
                                { Auth-Session-State }
                                { Origin-Host }
                                { Origin-Realm } 
```

Would it be an option to rename the old one or is it now set in stone?

fiorix:
Good catch @dunst0, you're right — the 3GPP specs (TS 29.229, TS 29.273) define it as Multimedia-Auth-Request/Answer (MAR/MAA), not Multimedia-Authentication.

Fixed in ce954f4 — renamed the SWx dictionary entry to `Multimedia-Auth` and the Go constant to `MultimediaAuth`. A deprecated `MultimediaAuthentication` alias is kept for backward compatibility.

PR #224, opened 2026-04-21
Fix unknown command code crashes connection (RFC 6733 §6.2)

### LINKED ISSUE #223: Unknown command handling is incorrect
Se below

tridentsx:
## Summary

When a message arrives with a command code not in the loaded dictionary, the server hard-closes the TCP connection. Per RFC 6733 §6.2, the correct behavior is to respond with Result-Code **3001 (DIAMETER_COMMAND_UNSUPPORTED)** with the **E-bit** set, and keep the connection open.

Fixes #223

## Changes

- **`readHeader`**: return nil `cmd` without error on unknown command so the message body is still drained from the socket
- **`readBody`**: handle nil `cmd` by decoding AVPs best-effort
- **`ReadMessage`**: return `ErrCommandUnsupported` sentinel with the fully parsed `*Message` so callers can build a proper answer
- **`conn.serve`**: catch `ErrCommandUnsupported` on requests, auto-respond with 3001 + E-bit, continue the read loop

## Testing

New test file `diam/unknown_avp_test.go` covers:
- Unknown command → receives 3001 answer with E-bit, correct command code echoed, connection stays alive, follow-up messages work
- Unknown AVPs (M=0, M=1, vendor-specific, inside grouped) — connection survives in all cases
- Raw bytes with non-standard vendor AVPs decoded correctly

All existing tests pass with no regressions (`go test ./...`).

### DIFF
--- a/diam/message.go
+++ b/diam/message.go
@@ -63,12 +63,12 @@ func readerBufferSlice(buf *bytes.Buffer, l int) []byte {
 	return make([]byte, l)
 }
 
-// ReadMessage reads a binary stream from the reader and uses the given
 // ErrCommandUnsupported is returned by ReadMessage when the message
 // contains an unknown command code. The Message is fully parsed so
 // the caller can build a proper 3001 answer per RFC 6733 §6.2.
 var ErrCommandUnsupported = errors.New("diameter: command unsupported")
 
+// ReadMessage reads a binary stream from the reader and uses the given
 // dictionary to parse it.
 func ReadMessage(reader io.Reader, dictionary *dict.Parser) (*Message, error) {
 	buf := newReaderBuffer()
--- a/diam/message.go
+++ b/diam/message.go
@@ -64,6 +64,11 @@ func readerBufferSlice(buf *bytes.Buffer, l int) []byte {
 }
 
 // ReadMessage reads a binary stream from the reader and uses the given
+// ErrCommandUnsupported is returned by ReadMessage when the message
+// contains an unknown command code. The Message is fully parsed so
+// the caller can build a proper 3001 answer per RFC 6733 §6.2.
+var ErrCommandUnsupported = errors.New("diameter: command unsupported")
+
 // dictionary to parse it.
 func ReadMessage(reader io.Reader, dictionary *dict.Parser) (*Message, error) {
 	buf := newReaderBuffer()
@@ -78,6 +83,10 @@ func ReadMessage(reader io.Reader, dictionary *dict.Parser) (*Message, error) {
 	m.stream = stream
 	if err = m.readBody(reader, buf, cmd, stream); err != nil {
 		return m, err
+	}
+	if cmd == nil {
+		// Header decoded OK but command not in dictionary.
+		return m, ErrCommandUnsupported
 	}
 	if dictionary.Strict {
 		return m, m.DecodeErr
@@ -113,7 +122,9 @@ func (m *Message) readHeader(r io.Reader, buf *bytes.Buffer) (cmd *dict.Command,
 		m.Header.CommandCode,
 	)
 	if err != nil {
-		return nil, stream, err
+		// Unknown command: return nil cmd (no IO error) so the
+		// body can still be drained and an answer sent per RFC 6733.
+		return nil, stream, nil
 	}
 	return cmd, stream, nil
 }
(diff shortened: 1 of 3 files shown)


fiorix (changes requested):
Good fix for a real problem. RFC 6733 compliance is correct in direction. Three items to address.

## 1. `ErrCommandUnsupported` splits `ReadMessage` godoc (bug)

The var declaration is inserted between lines 1 and 2 of the `ReadMessage` doc comment block. Result:

```go
// ReadMessage reads a binary stream from the reader and uses the given
// ErrCommandUnsupported is returned by ReadMessage when ...
var ErrCommandUnsupported = errors.New("diameter: command unsupported")

// dictionary to parse it.
func ReadMessage(...
```

`ReadMessage`'s godoc becomes just `"dictionary to parse it."` and `ErrCommandUnsupported`'s comment merges with the first line of the old `ReadMessage` doc. Move the var + its comment above the `ReadMessage` doc block or to a separate location.

## 2. `WriteTo` error ignored in `conn.serve()` (bug)

```go
ans.WriteTo(c.writer)
continue
```

If the write fails (broken pipe, etc.), the loop continues and the next `readMessage` will likely fail too, but silently. Handle the error:

```go
if _, err := ans.WriteTo(c.writer); err != nil {
    break
}
```

## 3. Unknown answer closes connection

When an unknown-command answer arrives (R-bit=0), the code falls through to `c.rwc.Close()`. This is arguably too aggressive: unknown answers should be discarded (logged), not kill the connection. Consider:

```go
if errors.Is(err, ErrCommandUnsupported) && m != nil {
    if m.Header.CommandFlags&RequestFlag == RequestFlag {
        ans := m.Answer(CommandUnsupported)
        ans.Header.CommandFlags |= ErrorFlag
        if _, err := ans.WriteTo(c.writer); err != nil {
            break
        }
    }
    // Unknown answer: discard silently.
    continue
}
```

Tests are solid; good coverage of edge cases.

tridentsx:
Addressed all three points in cd27823.

### 1. `ErrCommandUnsupported` split `ReadMessage` godoc — fixed ✅

Moved the var + its doc above `ReadMessage`'s doc block so each declaration has its own clean godoc.

### 2. `WriteTo` error ignored — fixed ✅

The answer write path now checks the error and breaks the read loop on failure instead of continuing into a guaranteed-failing `readMessage`.

### 3. Unknown answers close the connection + RFC compliance gap — reworked ✅

Taking this further than a log-and-discard for answers. The previous revision was emitting a **non-compliant** 3001 answer because it did not include Origin-Host / Origin-Realm, which RFC 6733 §7.2 marks as required in `<answer-message>`. That would potentially cause strict peers to reject our answer.

Reworked the flow to be hybrid compliant-or-close:

```go
if errors.Is(err, ErrCommandUnsupported) && m != nil {
    if m.Header.CommandFlags&RequestFlag == RequestFlag {
        if c.server.OriginHost != "" && c.server.OriginRealm != "" {
            // Build RFC-compliant 3001: same Command Code, R cleared,
            // E set, Result-Code, Origin-Host, Origin-Realm, H2H echoed.
            ans := m.Answer(CommandUnsupported)
            ans.Header.CommandFlags |= ErrorFlag
            ans.NewAVP(avp.OriginHost, avp.Mbit, 0, c.server.OriginHost)
            ans.NewAVP(avp.OriginRealm, avp.Mbit, 0, c.server.OriginRealm)
            if _, werr := ans.WriteTo(c.writer); werr != nil {
                break
            }
            continue
        }
        // No identity: fall through to close (historical behavior).
    } else {
        log.Printf("diam: discarding unknown answer (cmd=%d) from %s", ...)
        continue
    }
}
c.rwc.Close()
```

Added `Server.OriginHost` / `Server.OriginRealm` (`datatype.DiameterIdentity`) so the low-level server can construct compliant answers. Empty (zero value) preserves the old close-on-error behavior, so existing users don't change.

### Tests

- `TestUnknownCommand` — server configured with identity, asserts 3001 answer has Origin-Host, Origin-Realm, E-bit, correct Result-Code, matching Hop-by-Hop.
- `TestUnknownCommand_Answer` — unknown answer logged and discarded, connection stays open.
- `TestUnknownCommand_NoIdentityClosesConn` — no identity configured ⇒ server closes the connection; peer sees EOF on next read.

All tests pass with `-race`.

PR #225, opened 2026-04-21
Enable SCTP_NODELAY to fix catastrophic SCTP performance

### LINKED ISSUE #175: Performance on sctp is very poor
hi,
this is not an issue just an info request.

Has anybody tried the example client/server on sctp?
Just because the sctp performance is very poor:
With tcp: 259058/s diameter message can be reached (on localhost)
With sctp only 6938/s message.

I opened an issue regarding the sctp package (https://github.com/ishidawataru/sctp/issues/60) but maybe somebody has a solution or workaround.
Or just I did something wrong....

I appreciate any hint or suggestion.
thanks

tridentsx:
## Summary

SCTP throughput is ~2,340x slower than TCP (#175). The root cause is that `SCTP_NODELAY` is never set on the socket, so the kernel's Nagle-like algorithm buffers small Diameter messages for up to ~200ms before sending.

Fixes #175

## Change

One line in `NewSCTPConn()` — set `SCTP_NODELAY` on every SCTP connection (both client and server side):

```go
noDelay := 1
sctpConn.Setsockopt(sctp.SCTP_NODELAY, uintptr(unsafe.Pointer(&noDelay)), unsafe.Sizeof(noDelay))
```

## Benchmark results

| Transport | Before | After |
|---|---|---|
| TCP | 81,492 ns/op | 90,596 ns/op |
| SCTP | 190,607,597 ns/op | **155,590 ns/op** |
| SCTP/TCP ratio | 2,340x slower | **1.7x slower** |

**1,225x improvement** — from minutes-per-thousand-messages to near TCP parity.

Includes a transport benchmark (`BenchmarkDiamTransport_TCP` / `BenchmarkDiamTransport_SCTP`) for ongoing comparison.

All existing tests pass.

### DIFF
--- a/diam/network_sctp.go
+++ b/diam/network_sctp.go
@@ -110,6 +110,9 @@ func NewSCTPConn(sctpConn *sctp.SCTPConn) MultistreamConn {
 		return nil
 	}
 	sctpConn.SubscribeEvents(sctp.SCTP_EVENT_DATA_IO)
+	// Disable Nagle-like buffering so small Diameter messages are sent immediately.
+	noDelay := 1
+	sctpConn.Setsockopt(sctp.SCTP_NODELAY, uintptr(unsafe.Pointer(&noDelay)), unsafe.Sizeof(noDelay))
 	return &SCTPConn{SCTPConn: sctpConn, s: &streams{}, currStream: InvalidStreamID, writerStream: InvalidStreamID}
 }
 
@@ -389,13 +392,14 @@ func (d sctpSingleStreamDialer) Dial(network, address string) (net.Conn, error) 
 		return nil, err
 	}
 
-	return sctp.DialSCTPExt(
+	conn, err := sctp.DialSCTPExt(
 		network,
 		d.LocalAddr,
 		sctpAddr,
 		sctp.InitMsg{
 			NumOstreams:  MaxOutboundSCTPStreams,
 			MaxInstreams: MaxInboundSCTPStreams})
+	return NewSCTPConn(conn), err
 }
 
 // Accept implements the Accept method in the listener interface for sctpListener (see: MultistreamListen).
--- a/diam/bench_transport_test.go
+++ b/diam/bench_transport_test.go
@@ -1,0 +1,58 @@
+package diam
+
+import (
+	"net"
+	"testing"
+
+	"github.com/fiorix/go-diameter/v4/diam/avp"
+	"github.com/fiorix/go-diameter/v4/diam/datatype"
+	"github.com/fiorix/go-diameter/v4/diam/dict"
+)
+
+func benchmarkDiamTransport(b *testing.B, network string) {
+	done := make(chan struct{}, 1)
+	mux := NewServeMux()
+	mux.HandleFunc("ALL", func(c Conn, m *Message) {
+		done <- struct{}{}
+	})
+
+	ln, err := MultistreamListen(network, "127.0.0.1:0")
+	if err != nil {
+		b.Skipf("cannot listen on %s: %v", network, err)
+	}
+	defer ln.Close()
+
+	srv := &Server{Handler: mux}
+	go srv.Serve(ln)
+
+	// Use the same dialer the library uses internally
+	dialer := getMultistreamDialer(network, 0, nil)
+	rwc, err := dialer.Dial(network, ln.Addr().String())
+	if err != nil {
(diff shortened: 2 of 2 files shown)


diam/network_sctp.go
```diff
@@ -110,6 +110,9 @@ func NewSCTPConn(sctpConn *sctp.SCTPConn) MultistreamConn {
 		return nil
 	}
 	sctpConn.SubscribeEvents(sctp.SCTP_EVENT_DATA_IO)
+	// Disable Nagle-like buffering so small Diameter messages are sent immediately.
+	noDelay := 1
+	sctpConn.Setsockopt(sctp.SCTP_NODELAY, uintptr(unsafe.Pointer(&noDelay)), unsafe.Sizeof(noDelay))
```
gomaja, line 115: I would suggest the below line of code instead.
It is perhaps better to leave unsafe pointer access to the dependency itself.
```suggestion
	// Disable Nagle-like buffering so small Diameter messages are sent immediately.
	sctpConn.SetSackTimer(&sctp.SackTimer{SackDelay: 0, SackFrequency: 1})
```

tridentsx: Thanks for the review! I agree that avoiding `unsafe` is preferable when a higher-level API exists. However, `SetSackTimer`controls the **receiver-side** SACK delay (RFC 6458 §8.1.19), while `SCTP_NODELAY` controls the **sender-side** Nagle algorithm (RFC 6458 §8.1.21) — they're different socket options.
 
The performance problem is caused by sender-side buffering (Nagle), so `SCTP_NODELAY` is the correct fix. Unfortunately the `ishidawataru/sctp` package doesn't expose a `SetNoDelay()` helper, so `Setsockopt` with `unsafe.Pointer` is the only way to set it.

That said, setting `SackDelay: 0` on the receiver side could be a useful *additional* optimization. Happy to add both if you think it's worth it

fiorix (changes requested):
The SCTP_NODELAY fix is correct and well-motivated. However, this PR bundles two unrelated changes that should be separate.

## 1. Split the PR (blocking)

The `message.go`, `server.go`, and `unknown_avp_test.go` changes are identical to PR #224 (unknown command handling). This PR should contain only the SCTP_NODELAY fix and the transport benchmark. The unknown-command work should land via #224 alone.

## 2. `sctpSingleStreamDialer.Dial` misses NODELAY (bug)

`sctpSingleStreamDialer.Dial` (line 386) calls `sctp.DialSCTPExt` and returns the raw `*sctp.SCTPConn` directly, bypassing `NewSCTPConn`. Connections made through this path won't get SCTP_NODELAY. Either apply the setsockopt there too, or refactor to route through `NewSCTPConn`.

## 3. `Setsockopt` error dropped (minor)

The existing `SubscribeEvents` also drops its error, so this matches the pattern. Worth a `log.Printf` on failure for diagnostics, but not a blocker.

Transport benchmarks are a nice addition.

diam/network_sctp.go
tridentsx: Sorry I mixed up the commit , it should be fixed now

fiorix (approved):
Both issues addressed:

1. Unrelated unknown-command code removed -- PR is now SCTP-only
2. `sctpSingleStreamDialer.Dial` now routes through `NewSCTPConn`, matching the pattern already used in `sctpListener.Accept`

LGTM.

diam/network_sctp.go
gomaja: > 
> 
>   Thanks for the review! I agree that avoiding `unsafe` is preferable when a higher-level API exists. However, `SetSackTimer`controls the **receiver-side** SACK delay (RFC 6458 §8.1.19), while `SCTP_NODELAY` controls the **sender-side** Nagle algorithm (RFC 6458 §8.1.21) — they're different socket options.
> 
>  
> 
> The performance problem is caused by sender-side buffering (Nagle), so `SCTP_NODELAY` is the correct fix. Unfortunately the `ishidawataru/sctp` package doesn't expose a `SetNoDelay()` helper, so `Setsockopt` with `unsafe.Pointer` is the only way to set it.
> 
> 
> 
> That said, setting `SackDelay: 0` on the receiver side could be a useful *additional* optimization. Happy to add both if you think it's worth it

Yes you are right. 
Then it will also be beneficial for having both, it is worth it.

PR #226, opened 2026-04-21
Fix Unmarshal ignoring VendorId for AVPs with same code (#169)

### LINKED ISSUE #169: unmarshal issue when two avp's with same id but different VendorId used
It seems the code does not support the case when the same message has two AVPs that have the same avp-id but different VendorIds. I have run into this with the User-Password and 3GPP-Charging-Id used in AAR message - both of these have the avp-id of 2 but the vendorId is different. 

It seems Marshal seems to include the VendorId into the message but unmarshal ignores the VendorId so the data of User-Password gets copied into both the avps. 

I did find a comment like this:
		// Relies on the fact that in the same app will not be AVPs with same code but different vendorId

but not sure how it is ok to make such an assumption and why is it necessary. Is it NOT possible to support this case or was it a temporary shortcut? Would it be possible to fix this issue?

tridentsx:
Fixes #169

## Problem

`newIndex()` in `reflect.go` keys AVPs by `Code` only. When two AVPs share the same code but have different VendorIds (e.g. User-Password code=2 vendor=0 and 3GPP-Charging-Id code=2 vendor=10415), they end up in the same index bucket. `scanStruct()` then picks `avps[0]` regardless of VendorId, causing the wrong data to be assigned.

## Fix

Key the index by `{Code, VendorID}` instead of just `Code`. The lookup in `scanStruct` now uses both the dictionary AVP code and vendor ID.

## Testing

New test `TestUnmarshalVendorId` confirms AVPs with same code but different VendorIds are correctly separated. All existing tests pass.

### DIFF
--- a/diam/reflect.go
+++ b/diam/reflect.go
@@ -388,7 +388,7 @@ func scanStruct(m *Message, field reflect.Value, avps []*AVP) error {
 		}
 		// Lookup the AVP name (tag) in the dictionary.
 		// The dictionary AVP has the code.
-		d, err := m.Dictionary().FindAVP(m.Header.ApplicationID, avpname) // Relies on the fact that in the same app will not be AVPs with same code but different vendorId
+		d, err := m.Dictionary().FindAVP(m.Header.ApplicationID, avpname)
 		if err != nil {
 			return err
 		}
--- a/diam/reflect.go
+++ b/diam/reflect.go
@@ -349,12 +349,18 @@ func (m *Message) Unmarshal(dst interface{}) error {
 	return scanStruct(m, v, m.AVP)
 }
 
-// newIndex returns a map of AVPs indexed by their code.
-// TODO: make this part of the Message.
-func newIndex(avps []*AVP) map[uint32][]*AVP {
-	idx := make(map[uint32][]*AVP, len(avps))
+// avpKey identifies an AVP by both Code and VendorID.
+type avpKey struct {
+	Code     uint32
+	VendorID uint32
+}
+
+// newIndex returns a map of AVPs indexed by their {Code, VendorID}.
+func newIndex(avps []*AVP) map[avpKey][]*AVP {
+	idx := make(map[avpKey][]*AVP, len(avps))
 	for _, a := range avps {
-		idx[a.Code] = append(idx[a.Code], a)
+		key := avpKey{a.Code, a.VendorID}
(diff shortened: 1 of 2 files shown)


fiorix (changes requested):
Correct fix for a real bug (#169). The `{Code, VendorID}` composite key is the right approach. Two items to address.

## 1. Stale comment on line 385

```go
d, err := m.Dictionary().FindAVP(m.Header.ApplicationID, avpname) // Relies on the fact that in the same app will not be AVPs with same code but different vendorId
```

This comment describes the old assumption that this PR fixes. Remove it.

## 2. Tests only cover `newIndex`, not the full `Unmarshal` path

Both tests exercise `newIndex` directly but never call `m.Unmarshal(&struct{})` through `scanStruct`. Add an end-to-end test: define a struct with tags for two AVPs that share a code but differ by vendor (e.g. `avp:"User-Password"` and a vendor-specific AVP), call `m.Unmarshal`, and verify both fields get the right data.

The two existing tests are also near-identical and should be consolidated into one.

tridentsx:
Both points fixed in bf53f7c.

### 1. Stale comment on the `FindAVP` line — fixed ✅

Removed. The `{Code, VendorID}` index key introduced by the PR makes that assumption unnecessary.

### 2. Tests only exercised `newIndex`, not `Unmarshal` — fixed ✅

Replaced the two near-identical `newIndex`-level tests with a single end-to-end test driving the public `Unmarshal` API. Self-contained: loads an inline dictionary defining two AVPs with code 777 — `Example-Base` (no vendor) and `Example-Vendor` (vendor-id 10415) — then builds a message with both, unmarshals into a struct, and asserts both fields get the correct data.

```go
var dst struct {
    Base   datatype.OctetString `avp:"Example-Base"`
    Vendor datatype.Unsigned32  `avp:"Example-Vendor"`
}
if err := m.Unmarshal(&dst); err != nil {
    t.Fatalf("Unmarshal: %v", err)
}
// Asserts dst.Base == "base-data" and dst.Vendor == 12345
```

Swap the two AVPs in the message builder and the test fails — so it genuinely catches the bug the PR fixes, and would also catch any future regression in `scanStruct`'s use of the index, not just in `newIndex`.

Net effect: 2 files changed, +37 / −68 lines. `go test -race ./...` — all packages pass.

PR #227, opened 2026-04-21
Fix E.164 Address decode: strip family prefix consistently

### LINKED ISSUE #201: Trying to Decode E164 Address Type But its failing
Hi ,
Can some one help us to understand why the decode for Address Family is not happening as expected. 

We have SMSC-Address , which is of Address Family and of type E164, when we define "Address" in our dictionary and try to Decode this it adds the Type  value at the front of the parsed value as this  "\x00\b436990071301" instead of "436990071301".

SMSC_Addres

Example Parsed RAW value: 

SMSC-Address {Code:2017,Flags:0xc0,Length:28,VendorId:10415,Value:Address{datatype.Address{0x34, 0x33, 0x36, 0x39, 0x39, 0x30, 0x30, 0x37, 0x31, 0x33, 0x30, 0x31}}, Type{datatype.Address{0x0, 0x8}} Padding:2},

func TestDecodeAddressE164(t *testing.T) {
	b := []byte{0x00, 0x08, 0x34, 0x33, 0x36, 0x39, 0x39, 0x30, 0x30, 0x37, 0x31, 0x33, 0x30, 0x31}
	address, err := DecodeAddress(b)
	if err != nil {
		t.Fatal(err)
	}
	if address.Padding() != 3 {
		t.Fatalf("Unexpected padding. Want 3, have %d",
			address.Padding())
	}
}

Result : --- FAIL: TestDecodeAddressE164 (1.87s)
    address_test.go:206: Unexpected padding. Want 3, have 2
FAIL

tridentsx:
Fixes #201

## Problem

`DecodeAddress` strips the 2-byte address family prefix for IPv4 (0x0001) and IPv6 (0x0002) but returns the **full bytes including the prefix** for other types like E.164 (0x0008). This causes E.164 addresses to include the raw family bytes in the decoded value:

```
"\x00\b436990071301"  // wrong — includes 0x00 0x08 prefix
"436990071301"        // correct — just the phone number
```

## Fix

Strip the family prefix for all address types in `DecodeAddress`, not just IPv4/IPv6. Update `Serialize` to re-add the E.164 family prefix (0x0008) during serialization, and fix `Len`/`Padding` to consistently account for the 2-byte prefix.

Round-trip test confirms: decode → serialize produces the original wire bytes.

All existing tests pass.

### DIFF
--- a/diam/datatype/address.go
+++ b/diam/datatype/address.go
@@ -34,8 +34,11 @@ func DecodeAddress(b []byte) (Type, error) {
 			return Address{}, errors.New("Invalid length for IPv6")
 		}
 	default:
+		// Preserve the full slice (family prefix + value) for unknown
+		// address families so Serialize can round-trip them untouched.
 		return Address(d), nil
 	}
+	// IPv4/IPv6: strip the 2-byte address family prefix.
 	return Address(d[2:]), nil
 }
 
@@ -51,6 +54,7 @@ func (addr Address) Serialize() []byte {
 		b[1] = 0x02
 		copy(b[2:], addr)
 	} else {
+		// Non-IP: addr already contains the family prefix from decode.
 		b = make([]byte, len(addr))
 		copy(b, addr)
 	}
@@ -60,24 +64,16 @@ func (addr Address) Serialize() []byte {
 // Len implements the Type interface.
 func (addr Address) Len() int {
 	if ip4 := net.IP(addr).To4(); ip4 != nil {
-		return len(ip4) + 2 // two bytes from the address family
+		return len(ip4) + 2
 	} else if ip6 := net.IP(addr).To16(); ip6 != nil {
-		return len(ip6) + 2 // two bytes from the address family
-	} else {
-		return len(addr)
+		return len(ip6) + 2
 	}
+	return len(addr) // family prefix is already in the slice
 }
 
 // Padding implements the Type interface.
 func (addr Address) Padding() int {
-	var l int
-	if ip4 := net.IP(addr).To4(); ip4 != nil {
-		l = len(ip4) + 2 // two bytes from the address family
-	} else if ip6 := net.IP(addr).To16(); ip6 != nil {
-		l = len(ip6) + 2 // two bytes from the address family
-	} else {
-		l = len(addr)
-	}
+	l := addr.Len()
 	return pad4(l) - l
 }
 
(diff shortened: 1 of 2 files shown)


fiorix (changes requested):
Addresses a real bug (#201) but the implementation has two problems: one causes data corruption, the other causes a panic.

## 1. `Serialize()` hardcodes E.164 family for all non-IP addresses (data corruption)

The else branch unconditionally writes `b[1] = 0x08`:

```go
b = make([]byte, 2+len(addr))
b[1] = 0x08
copy(b[2:], addr)
```

IANA defines 15+ address families (NSAP=3, HDLC=4, E.163=7, E.164=8, F.69=9, X.121=10, IPX=11, ...). If a peer sends family 3 (NSAP), `DecodeAddress` strips the prefix, then `Serialize()` re-encodes it as E.164 (0x0008). The round-trip invariant breaks for any non-IP, non-E.164 family.

The root cause is that `Address` is `[]byte` with no room to store the family. The minimal fix is to keep the family prefix in the slice for non-IP types (current master behavior) and only fix `Len()`/`Padding()`/`String()` to account for it correctly. This resolves #201 without breaking round-trip for any address family.

## 2. `pretty_dump.go:221` not updated (panic)

```go
return fmt.Sprintf("%#v, %#v", addr[2:], addr[:2])
```

After this PR, decoded non-IP `Address` values no longer contain the family prefix. This line slices into the actual data, and will panic on addresses shorter than 2 bytes. Must update to match the new representation, or this becomes moot if you take the minimal-fix approach from item 1.

tridentsx:
Good catches, both fixed in the force-pushed 46be60d.

### Approach change

Rebased the branch and reworked the commit per your minimal-fix recommendation. Instead of stripping the family prefix on decode (which required re-adding 0x0008 on serialize and lost the family identity for non-E.164 non-IP types), I kept master's representation — the 2-byte family prefix stays in the slice for non-IP addresses — and only changed what was actually broken for #201.

### 1. Hardcoded E.164 on Serialize — avoided

`DecodeAddress` and `Serialize` for non-IP now match master exactly. Round-trip is preserved for any IANA address family (NSAP, HDLC, IPX, ...): the family bytes are kept untouched in the slice on decode and written back verbatim on serialize.

### 2. `pretty_dump.go:221` — no change needed

Since non-IP addresses still carry the family prefix in the slice, `addr[2:]` and `addr[:2]` in `pretty_dump.go:221` still point at the correct bytes. That file needs no changes.

### What actually fixes #201

Only `Address.String()` changed: the non-IP branch now renders the value as a readable string (via `string(addr[2:])`) instead of the raw byte repr. With an added guard for `len(addr) < 2` to avoid slicing past the end. Before/after:

```
before: Address{[]byte{0x34,0x33,0x36,...}}, Type{[]byte{0x0,0x8}} Padding:2
after:  Address{436990071301},Type{[]byte{0x0, 0x8}},Padding:2
```

### Tests

`TestDecodeAddressE164` expanded with:
- value-after-prefix assertion
- `Len()` assertion
- wire-level round-trip assertion
- regression guard that `String()` never embeds raw `\x00\x08` into the rendered value

### Verification

`go test -race -count=1 ./...` — all packages pass.

Net diff vs master: +39 / −18 across 2 files.

fiorix (approved):
Approach change is sound: preserving the family prefix for non-IP types instead of strip-and-reconstruct eliminates the magic-byte concern entirely. Round-trip is correct, tests are solid.

Minor nit for a follow-up: `pretty_dump.go:221` still uses `%#v` and lacks the `len < 2` guard that `String()` now has. Not a blocker -- pre-existing wart, not a regression.

PR #228, opened 2026-04-21
Dispatch message handlers concurrently in conn.serve()

### LINKED ISSUE #139: When TPS is high, mux handler dropped messages.
Hi, dear go-diameter author:
    Thanks for your code, and I used it to build a good diameter client program.
    But I found, when I send requests in a high rate(above 50 entries per seconds), and the anwser came back also in that rate. Some answers were dropped.
    I just write a few code in answers handler, so it should not comsume much CPU time (only a number++), I guess the buffer in your code is too short for buffering the incoming answers.
    Is there any possible to adjust that?
    Wait for your response.
    Thanks.

BR
ZZ

tridentsx:
Fixes #139, #195

## Problem

`conn.serve()` processes messages sequentially — a slow handler blocks all subsequent reads on the connection. Under high load, the kernel TCP buffer fills and messages are effectively lost.

## Fix

Two changes to `server.go`:
- **`go` dispatch**: `go serverHandler{...}.ServeDIAM(c.writer, m)` instead of blocking call
- **`notifyClientGone()` on exit**: ensures `CloseNotify` fires correctly when a handler goroutine closes the connection (the pipe swap in `liveSwitchReader` may not have been set up yet when the read loop exits)

One change to `examples/s6a_proxy/service/ai.go`:
- Use `m.MessageStream()` instead of `msc.CurrentStream()` since the connection's current stream may change between dispatch and handler execution

## Why this is safe

Diameter matches answers to requests by **Hop-by-Hop ID**, not arrival order. Concurrent dispatch does not affect clients — same as HTTP/2 multiplexing.

`response.Write` already has a `sync.Mutex`, so concurrent handler goroutines writing responses is safe.

## Testing

All existing tests pass including SCTP stream tests. Race detector clean.

### DIFF
--- a/diam/concurrent_serve_test.go
+++ b/diam/concurrent_serve_test.go
@@ -12,8 +12,8 @@ import (
 )
 
 // cerPayload builds a minimal CER payload usable by the test mux.
-func cerPayload(t *testing.T) []byte {
-	t.Helper()
+func cerPayload(tb testing.TB) []byte {
+	tb.Helper()
 	msg := NewRequest(257, 0, dict.Default)
 	msg.NewAVP(avp.OriginHost, avp.Mbit, 0, datatype.DiameterIdentity("test"))
 	msg.NewAVP(avp.OriginRealm, avp.Mbit, 0, datatype.DiameterIdentity("test"))
@@ -22,7 +22,7 @@ func cerPayload(t *testing.T) []byte {
 	msg.NewAVP(avp.ProductName, 0, 0, datatype.UTF8String("test"))
 	b, err := msg.Serialize()
 	if err != nil {
-		t.Fatalf("serialize: %v", err)
+		tb.Fatalf("serialize: %v", err)
 	}
 	return b
 }
@@ -224,7 +224,7 @@ func runDispatchBenchmark(b *testing.B, maxConcurrent int, handlerLatency time.D
 			ln.Close()
 			b.Fatal(err)
 		}
-		payload := cerPayload(&testing.T{})
+		payload := cerPayload(b)
 
 		b.StartTimer()
 		for j := 0; j < total; j++ {
--- a/diam/concurrent_serve_test.go
+++ b/diam/concurrent_serve_test.go
@@ -1,0 +1,271 @@
+package diam
+
+import (
+	"net"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/fiorix/go-diameter/v4/diam/avp"
+	"github.com/fiorix/go-diameter/v4/diam/datatype"
+	"github.com/fiorix/go-diameter/v4/diam/dict"
+)
+
+// cerPayload builds a minimal CER payload usable by the test mux.
+func cerPayload(t *testing.T) []byte {
+	t.Helper()
+	msg := NewRequest(257, 0, dict.Default)
+	msg.NewAVP(avp.OriginHost, avp.Mbit, 0, datatype.DiameterIdentity("test"))
+	msg.NewAVP(avp.OriginRealm, avp.Mbit, 0, datatype.DiameterIdentity("test"))
+	msg.NewAVP(avp.HostIPAddress, avp.Mbit, 0, datatype.Address(net.ParseIP("127.0.0.1")))
+	msg.NewAVP(avp.VendorID, avp.Mbit, 0, datatype.Unsigned32(0))
+	msg.NewAVP(avp.ProductName, 0, 0, datatype.UTF8String("test"))
+	b, err := msg.Serialize()
+	if err != nil {
+		t.Fatalf("serialize: %v", err)
+	}
+	return b
+}
+
+// TestConcurrentServeThroughput verifies that enabling MaxConcurrentHandlers
+// actually dispatches handlers concurrently. With a 1ms handler and 1000
+// messages, sequential dispatch would take >= 1s; concurrent should be
+// dramatically faster.
+func TestConcurrentServeThroughput(t *testing.T) {
+	var received int64
+	const total = 1000
+
+	mux := NewServeMux()
+	mux.HandleFunc("ALL", func(c Conn, m *Message) {
+		time.Sleep(1 * time.Millisecond)
+		atomic.AddInt64(&received, 1)
+	})
+
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer ln.Close()
+	srv := &Server{Handler: mux, MaxConcurrentHandlers: 256}
+	go srv.Serve(ln)
+
+	conn, err := net.Dial("tcp", ln.Addr().String())
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer conn.Close()
+
+	payload := cerPayload(t)
+	start := time.Now()
+	for i := 0; i < total; i++ {
+		if _, err := conn.Write(payload); err != nil {
+			t.Fatal(err)
+		}
+	}
+	for time.Since(start) < 5*time.Second {
+		if atomic.LoadInt64(&received) >= total {
+			break
+		}
+		time.Sleep(5 * time.Millisecond)
+	}
+	elapsed := time.Since(start)
+	got := atomic.LoadInt64(&received)
+	t.Logf("Sent %d, received %d in %v (%.0f msg/s)", total, got, elapsed, float64(got)/elapsed.Seconds())
+	if got < total {
+		t.Errorf("Lost %d messages", total-got)
+	}
+	if elapsed > 500*time.Millisecond {
+		t.Errorf("Too slow (%v) — handlers may still be sequential", elapsed)
+	}
+}
+
+// TestConcurrentServePanicRecovery verifies that a panic inside a
+// concurrently-dispatched handler does not crash the server and that
+// subsequent messages are still processed.
+func TestConcurrentServePanicRecovery(t *testing.T) {
+	var received int64
+
+	mux := NewServeMux()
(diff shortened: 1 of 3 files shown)


tridentsx:
## Benchmark: 1000 messages with 1ms handler

| | Time | Throughput |
|---|---|---|
| **Sequential** (before) | 1,210ms | 826 msg/s |
| **Concurrent** (after) | 14ms | 72,690 msg/s |
| **Improvement** | **88x faster** | |

With real-world handlers that do DB lookups, logging, or response building (typically 1–10ms), sequential dispatch caps throughput at `1/handler_latency` regardless of network speed.

fiorix (changes requested):
Right diagnosis: sequential dispatch under high TPS is a real problem (#139, #195). But the one-line `go` change has serious safety gaps.

## 1. Handler panics crash the process (regression, blocking)

The existing `recover()` in `conn.serve()`'s defer catches panics in the read loop. Before this PR, handler panics were also caught because handlers ran in the same goroutine. After this PR, a handler panic in a spawned goroutine is unrecovered and kills the process. Wrap the dispatch:

```go
go func() {
    defer func() {
        if err := recover(); err != nil {
            buf := make([]byte, 4096)
            buf = buf[:runtime.Stack(buf, false)]
            log.Printf("diam: panic serving %v: %v\n%s",
                c.rwc.RemoteAddr().String(), err, buf)
        }
    }()
    serverHandler{c.server}.ServeDIAM(c.writer, m)
}()
```

## 2. Unbounded goroutines (blocking)

Every inbound message spawns a goroutine with no limit. Under high TPS with slow handlers (the exact scenario this fixes), goroutines accumulate without bound, trading "TCP buffer full" for "OOM / goroutine explosion". Needs a bounded worker pool or semaphore, e.g.:

```go
// In Server struct:
MaxConcurrentHandlers int

// In conn.serve(), initialized from server config:
sem := make(chan struct{}, maxHandlers)
// ...
sem <- struct{}{}
go func() {
    defer func() { <-sem }()
    // recover + ServeDIAM
}()
```

## 3. Handler interface contract broken (breaking change)

The `Handler` doc (line 27-29) says:

> Returning signals that the request is finished and that the server can move on to the next request on the connection.

This implies sequential semantics. User code that maintains per-connection state without locks will now race. Update the doc and consider making concurrency opt-in (default `MaxConcurrentHandlers=1` preserves current behavior).

## 4. No drain on connection close (should fix)

When the read loop exits, `c.rwc.Close()` fires immediately. Handlers still running in goroutines may be mid-write. A `sync.WaitGroup` should gate the close until inflight handlers complete.

## 5. No new tests

Should add tests for: concurrent dispatch actually works, handler panic doesn't crash the process.

The `notifyClientGone()` fix and the example `MessageStream()` change are both correct.

tridentsx:
Thanks for the review — all 5 points were valid and I've addressed each in 42ccdfa. Summary below, keyed to your numbered items.

### 1. Handler panics crash the process — fixed ✅

Added a `recover()` inside each spawned handler goroutine that logs the panic with stack trace in the same format as the existing read-loop recover. Covered by the new `TestConcurrentServePanicRecovery`: it sends 3 messages, panics in the first, and asserts the server processes the remaining 2.

### 2. Unbounded goroutines — fixed ✅

Added a per-connection semaphore sized by `Server.MaxConcurrentHandlers`. When the cap is reached the read loop blocks until a slot frees — backpressure flows back to the TCP layer as intended, no goroutine explosion. Covered by `TestConcurrentServeBounded`, which blocks 20 handlers on a channel and asserts peak in-flight ≤ limit.

### 3. Handler interface contract broken — fixed ✅

- **Doc updated**: removed the sequential promise and made the concurrency contract explicit, tied to `MaxConcurrentHandlers`.
- **Opt-in concurrency**: added `Server.MaxConcurrentHandlers int`. I chose `0 = sequential (default)` rather than your suggested `1 = sequential`. Rationale:
  - Zero value maps to existing behaviour so current users recompile and get identical semantics.
  - The `0` path takes a fast lane that calls `ServeDIAM` directly on the read-loop goroutine — no goroutine spawn, no semaphore ops. With `1 = sequential` we'd still pay that overhead per message for no benefit.
  - Encoding: `0` sequential, `>0` bounded concurrent, `<0` unbounded (documented as not recommended for untrusted peers).

  Happy to switch to `1 = sequential` if you'd prefer the explicit encoding.

### 4. No drain on connection close — fixed ✅

Added a per-connection `sync.WaitGroup`; `conn.serve()`'s defer now does `hwg.Wait()` before `c.rwc.Close()`, so in-flight handlers finish their writes before the socket closes.

### 5. No new tests — fixed ✅

Added in `diam/concurrent_serve_test.go`:
- `TestConcurrentServeThroughput` — updated to enable `MaxConcurrentHandlers: 256`, verifies actual concurrent dispatch (~49× faster than sequential under 1ms handler).
- `TestConcurrentServePanicRecovery` — verifies handler panics don't crash the server.
- `TestConcurrentServeBounded` — verifies the semaphore cap.
- `BenchmarkDispatch{Sequential,Concurrent}{1ms,NoLatency}` — quantify the trade-off.

### Benchmark numbers (local, AMD Ryzen AI 7 PRO 350)

| Workload | Sequential | Concurrent (256) | Speedup |
|---|---:|---:|---:|
| 1ms handler | ~810 msg/s | ~39,900 msg/s | **~49×** |
| No-op handler | ~107,000 msg/s | ~158,000 msg/s | ~1.5× |

The no-op case shows the floor: when there's no latency to hide, concurrent mode only pays its goroutine+semaphore tax. The 1ms case shows the real-world win for handlers doing any I/O. This is why concurrency is opt-in — zero-overhead default, big gains when enabled.

All tests pass with `-race`. `notifyClientGone()` and the s6a_proxy `MessageStream()` change are unchanged per your approval.

fiorix (changes requested):
Good rework. `dispatch()` method, semaphore, WaitGroup drain, defer ordering, and sequential default are all correct. One significant issue remains.

## 1. SCTP writes are unprotected under concurrent dispatch (blocking)

The standard handler write path `m.WriteTo(c)` goes through `WriteToStreamWithRetry`, which type-switches on `MultistreamWriter`. Since `*response` always satisfies `MultistreamWriter`, this calls `response.WriteStream` -- which has no mutex. For TCP this falls through to the mutex-protected `response.Write`, but for SCTP it goes directly to `msc.WriteStream` with no synchronization.

The write path for SCTP:

```
m.WriteTo(c)
  -> WriteToStreamWithRetry(writer, stream, 0)
    -> case MultistreamWriter:        // *response always matches
      -> w.WriteStream(b, stream)     // response.WriteStream -- NO MUTEX
        -> msc.WriteStream(b, stream) // direct to sctp, no lock
```

This means the PR description's claim that "`response.Write` already has a `sync.Mutex`" doesn't apply to SCTP. The mutex is never reached on the SCTP write path.

Fix: add `w.mu.Lock()`/`defer w.mu.Unlock()` to `response.WriteStream`. Note the TCP fallback currently calls `w.Write(b)` which would double-lock `w.mu`, so it needs to be inlined or use an unlocked internal write method.

## 2. `cerPayload(&testing.T{})` in benchmarks (minor)

`runDispatchBenchmark` passes a zero-value `testing.T`. If `Serialize` fails, `Fatalf` calls `runtime.Goexit()` silently in the benchmark goroutine. Refactor `cerPayload` to accept `testing.TB` and pass `b` instead.

Everything else checks out: defer ordering (recover -> sem release -> wg.Done), `hwg.Add`/`Wait` sequencing, nil sem handling, `notifyClientGone` idempotency, test logic.

tridentsx:
Both points fixed in ea3e72e.

### 1. SCTP `WriteStream` unprotected — fixed ✅

You're right: `*response` always satisfies `MultistreamWriter`, so `WriteToStreamWithRetry` always routes through `response.WriteStream`, never `response.Write`. The TCP path was only accidentally safe because `WriteStream` fell through to `w.Write` which locks.

Refactored to a `writeLocked` helper so both public methods lock at entry and share the body without double-locking:

```go
func (w *response) Write(b []byte) (int, error) {
    w.mu.Lock()
    defer w.mu.Unlock()
    return w.writeLocked(b)
}

func (w *response) WriteStream(b []byte, stream uint) (int, error) {
    w.mu.Lock()
    defer w.mu.Unlock()
    if msc, isMulti := w.conn.rwc.(MultistreamConn); isMulti {
        return msc.WriteStream(b, stream)
    }
    return w.writeLocked(b)
}
```

SCTP writes are now serialized per connection. Race detector clean across the full suite.

### 2. `cerPayload(&testing.T{})` — fixed ✅

Changed signature to `cerPayload(tb testing.TB)` with `tb.Helper()` and `tb.Fatalf`. Benchmark now passes `b` directly.

### Benchmark re-run after the added lock

No measurable regression — the mutex is uncontended under the single-handler-per-message write pattern.

| Workload | Sequential | Concurrent (256) | Speedup |
|---|---:|---:|---:|
| 1ms handler | ~870 msg/s | ~81,300 msg/s | **~93×** |
| No-op handler | ~449k msg/s | ~362k msg/s | 0.8× (pure overhead, as expected) |

fiorix (approved):
All items from both review rounds are properly addressed. The SCTP WriteStream mutex fix is correct: locks w.mu, uses writeLocked for TCP fallback to avoid double-lock. Semaphore, WaitGroup drain, panic recovery, opt-in default -- all verified in the actual code.

One medium item for a follow-up: the pre-existing c.rwc.Close() inside the read-loop error path (server.go) fires before the defer c.hwg.Wait(), contradicting the comment about not writing to a closed connection. Low practical impact (connection is dead from read error), but the fix is straightforward: remove it from the loop and let the defer handle it exclusively.

PR #229, opened 2026-04-23
Reduce allocations in AVP decode and message encode hot paths

kulcsartibor:
---

### Files changed

**Committed:**
- `diam/avp.go`
- `diam/datatype/decoder.go`
- `diam/datatype/decoder_test.go`
- `diam/datatype/datatype.go`
- `diam/dict/util.go`
- `diam/dict/parser.go`
- `diam/group.go`
- `diam/header.go`
- `diam/uintconv.go`

---

### Optimizations implemented

**1. Typed `FindAVPByCode` — avoid `interface{}` boxing** (`diam/dict/util.go`, `diam/avp.go`)

Added `FindAVPByCode(appid, code, vendorID uint32)` as a typed fast-path for the hot decode loop. The existing `FindAVPWithVendor` takes `code interface{}`, which boxes the `uint32` on every call and requires a type switch. The decode path at `avp.go:95` now calls the typed method directly. The original `FindAVPWithVendor` is preserved for backward compatibility.

**2. Array-indexed decoder dispatch with map fallback** (`diam/datatype/decoder.go`)

Replaced the `map[TypeID]DecoderFunc` lookup in `Decode()` with a fixed-size `[maxTypeID]DecoderFunc` array. TypeIDs are small contiguous integers (0–18), so array indexing is a single bounds check vs. map hashing. `Decode()` falls back to the exported `Decoder` map for any TypeID the array doesn't cover, preserving the ability for external code to register custom decoders via `datatype.Decoder[myType] = myFunc`. The `Decoder` map now also includes `QoSFilterRuleType` and `IPv6Type` so it stays consistent with the array. Added test cases to verify the registration of custom TypeID and decoder function.

**3. Skip intermediate Grouped AVP copy** (`diam/avp.go`, `diam/group.go`)

Grouped AVPs were decoded twice: first `datatype.DecodeGrouped` copied all payload bytes into a `datatype.Grouped` (`make + copy`), then `diam.DecodeGrouped` immediately converted it back to `[]byte` to parse child AVPs. The new `DecodeGroupedFromBytes` takes the raw payload slice directly, eliminating the intermediate allocation. `DecodeFromBytes` now checks the dictionary type *before* dispatching, short-circuiting the grouped path. The original `DecodeGrouped` is preserved as a wrapper for backward compatibility.

**4. Sentinel errors with contextual wrapping** (`diam/avp.go`)

Replaced 5 `fmt.Errorf` calls on error paths in `DecodeFromBytes` with pre-allocated `errors.New` sentinel variables (`errAVPHeaderTooShort`, `errAVPDataTooShort`, `errAVPVendorTooShort`). Same for `errAVPSerializeNilData` in `Serialize`/`SerializeTo`. Removes formatting overhead and allocations on error paths, and reduces function code size which can help compiler inlining/optimization of the surrounding hot path. On the two `errAVPDataTooShort` return sites, the sentinel is wrapped with `fmt.Errorf("%w: have %d need %d", ...)` so debugging retains byte-count context while `errors.Is` identity is preserved.

**5. `putUint24` direct buffer write** (`diam/uintconv.go`, `diam/avp.go`, `diam/header.go`)

Added `putUint24(b []byte, n uint32)` (package-internal) that writes 3 big-endian bytes directly into a target buffer. Replaced all `copy(b, uint32to24(n))` calls in `AVP.SerializeTo` (1 call) and `Header.SerializeTo` (2 calls). Each `uint32to24` allocated a 3-byte slice that was immediately copied and discarded. This single change made `Header.SerializeTo` zero-allocation.

**6. Dictionary inheritance for apps with no own AVPs** (`diam/dict/parser.go`)

`mergeInheritedAVPs` originally built its app set only from `p.avpcode` keys, so applications declared in the dictionary with zero AVPs of their own (e.g. `<application id="1001" type="acct"/>`) were skipped and base AVPs were never inherited into their lookup index. `FindAVPByCode` would then miss and return `Unknown` AVPs, breaking CER parsing for such apps. The merge now also seeds the app set from `p.appcode`, so every declared app gets its ancestor chain processed at load time. Fixes `TestHandleCER_HandshakeMetadataTCP` and `TestHandleCER_HandshakeMetadata_CustomIP`, which were hanging on `<-sm.HandshakeNotify()`.

---

### Benchmark results (Apple M1 Max, 10 cores)

| Benchmark | Before | After | Δ ns/op | Δ allocs |
|---|---|---|---|---|
| DecodeAVP | 103 ns, 4 allocs, 80 B | 81 ns, 3 allocs, 72 B | -21.4% | -25% |
| ReadMessage | 1,557 ns, 47 allocs, 1,160 B | 1,144 ns, 31 allocs, 1,048 B | -26.5% | -34.0% |
| EncodeHeader | 17.1 ns, 1 alloc, 24 B | 2.5 ns, 0 allocs, 0 B | -85.4% | -100% |
| EncodeAVP | 39.7 ns, 2 allocs | 39.7 ns, 2 allocs | ~0% | 0% |
| WriteMessage | 473 ns, 14 allocs | 473 ns, 14 allocs | ~0% | 0% |

---

### Backward compatibility

All changes are backward-compatible:

- No exported types, functions, or method signatures were removed or changed.
- Original `FindAVPWithVendor`, `FindAVP`, `DecodeGrouped`, `uint32to24`, and the `Decoder` map are all preserved.
- `Decode()` still resolves custom decoders registered via `datatype.Decoder[myType] = myFunc` (fallback after the array miss).
- New exports added: `FindAVPByCode`, `DecodeGroupedFromBytes`.
- All existing tests pass (`go test ./...`).

### DIFF
--- a/diam/avp.go
+++ b/diam/avp.go
@@ -17,6 +17,15 @@ import (
 // Used to signal that parsing should not stop.
 type DecodeError error
 
+// Pre-allocated sentinel errors for the hot decode path.
+// Using errors.New avoids fmt.Errorf formatting overhead on error paths.
+var (
+	errAVPHeaderTooShort   = errors.New("not enough data to decode AVP header")
+	errAVPDataTooShort     = errors.New("not enough data to decode AVP")
+	errAVPVendorTooShort   = errors.New("not enough data to decode AVP with Vendor-ID")
+	errAVPSerializeNilData = errors.New("failed to serialize AVP: Data is nil")
+)
+
 // AVP is a Diameter attribute-value-pair.
 type AVP struct {
 	Code     uint32        // Code of this AVP
@@ -55,18 +64,17 @@ func DecodeAVP(data []byte, application uint32, dictionary *dict.Parser) (*AVP, 
 // It uses the given application id and dictionary for decoding the bytes.
 func (a *AVP) DecodeFromBytes(data []byte, application uint32, dictionary *dict.Parser) error {
 	if len(data) < 8 {
-		return fmt.Errorf("Not enough data to decode AVP header: %d bytes", len(data))
+		return fmt.Errorf("%w: have %d need %d", errAVPHeaderTooShort, len(data), 8)
 	}
 	a.Code = binary.BigEndian.Uint32(data[0:4])
 	a.Flags = data[4]
 	a.Length = int(uint24to32(data[5:8]))
 	if len(data) < a.Length {
-		return fmt.Errorf("Not enough data to decode AVP: %d != %d",
-			len(data), a.Length)
+		return fmt.Errorf("%w: have %d need %d", errAVPDataTooShort, len(data), a.Length)
 	}
 	data = data[:a.Length] // this cuts padded bytes off
 	if len(data) < 8 {
-		return fmt.Errorf("Not enough data to decode AVP header: %d bytes", len(data))
+		return fmt.Errorf("%w: have %d need %d", errAVPHeaderTooShort, len(data), 8)
 	}
 
 	var hdrLength int
@@ -74,7 +82,7 @@ func (a *AVP) DecodeFromBytes(data []byte, application uint32, dictionary *dict.
 	// Read VendorId when required.
 	if a.Flags&avp.Vbit == avp.Vbit {
 		if a.Length < 12 {
-			return fmt.Errorf("Not enough data to decode AVP with Vendor-ID: length %d < 12", a.Length)
+			return fmt.Errorf("%w: have %d need %d", errAVPVendorTooShort, a.Length, 12)
 		}
 		a.VendorID = binary.BigEndian.Uint32(data[8:12])
 		payload = data[12:]
@@ -84,29 +92,24 @@ func (a *AVP) DecodeFromBytes(data []byte, application uint32, dictionary *dict.
 		hdrLength = 8
 	}
 	// Find this code in the dictionary.
-	dictAVP, err := dictionary.FindAVPWithVendor(application, a.Code, a.VendorID)
+	dictAVP, err := dictionary.FindAVPByCode(application, a.Code, a.VendorID)
 	if err != nil && dictAVP == nil {
 		return err
 	}
 	bodyLen := a.Length - hdrLength
 	if n := len(payload); n < bodyLen {
-		return fmt.Errorf(
-			"Not enough data to decode AVP: %d != %d",
-			hdrLength, n,
-		)
+		return fmt.Errorf("%w: have %d need %d", errAVPDataTooShort, n, bodyLen)
 	}
-	a.Data, err = datatype.Decode(dictAVP.Data.Type, payload)
-	if err != nil {
-		return DecodeError(fmt.Errorf("%s(%d): %v", dictAVP.Name, dictAVP.Code, err))
-	}
-	// Handle grouped AVPs.
-	if a.Data.Type() == datatype.GroupedType {
-		a.Data, err = DecodeGrouped(
-			a.Data.(datatype.Grouped),
-			application, dictionary,
-		)
+	// Handle grouped AVPs directly to avoid an intermediate copy.
+	if dictAVP.Data.Type == datatype.GroupedType {
+		a.Data, err = DecodeGroupedFromBytes(payload, application, dictionary)
 		if err != nil {
 			return DecodeError(fmt.Errorf("%s(%d): Grouped{%v}", dictAVP.Name, dictAVP.Code, err))
+		}
+	} else {
+		a.Data, err = datatype.Decode(dictAVP.Data.Type, payload)
+		if err != nil {
+			return DecodeError(fmt.Errorf("%s(%d): %v", dictAVP.Name, dictAVP.Code, err))
 		}
 	}
 	return nil
@@ -116,7 +119,7 @@ func (a *AVP) DecodeFromBytes(data []byte, application uint32, dictionary *dict.
 // It requires at least the Code, Flags and Data fields set.
 func (a *AVP) Serialize() ([]byte, error) {
 	if a.Data == nil {
-		return nil, errors.New("Failed to serialize AVP: Data is nil")
+		return nil, errAVPSerializeNilData
 	}
 	b := make([]byte, a.Len())
 	err := a.SerializeTo(b)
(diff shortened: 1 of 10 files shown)


fiorix:
Thanks for the work on this. `main` has moved forward — five PRs merged today (#230, #231, #232, #233, #234), plus a new v4.2.1 tag. Could you rebase onto current `main` before marking ready for review? None of the recent changes touch the files in this PR so the rebase should be mechanical. Once it's rebased and marked ready, I'll review.

kulcsartibor:
I made few changes to the `decoder.go`. With migrating to an array lookup from map speeds up the parsing by around 45%. This has the drawback that internal parser cannot be overridden just by putting items to the `Decoder[]`. So created the `RegisterDecoder` method to allow built in type overrides too.

The `datatype.go` I have the`IPv6Type` at the end of the list, not nice but avoid breaking downstream code if someone has a hardcoded integer comparison: `if avp.Data.Type() == 10 { ... }`

fiorix:
Nice work, tests pass locally and the perf motivation is clear. A few things before I approve:

**1. `datatype.Decoder` semantic change is a breaking API shift for v4**
The last commit drops built-ins from the exported `Decoder` map entirely. Downstream code that iterates or reads from `Decoder` for built-in types now sees an empty slot. Two options:
- Keep `Decoder` pre-populated with built-ins (as before) and have `Decode` prefer `decoderArray`; slightly slower on map overrides but zero back-compat break.
- Keep the extension-only model but call it out in the PR description so I can add a CHANGELOG note and migration pointer to `RegisterDecoder`.

Prefer the first unless you have a reason against it.

**2. Split between `FindAVPByCode` (new) and `FindAVPWithVendor` (existing)**
`avp.go` uses the new typed fast path; `message.go` and `pretty_dump.go` still call `FindAVPWithVendor`. Fine for this PR, but worth a TODO or a follow-up to migrate; otherwise we carry two lookup paths indefinitely.

**3. Sentinel error wrapping is inconsistent**
`errAVPDataTooShort` is returned bare once and wrapped with `%w: have X need Y` twice. `errors.Is` works in both cases, but the asymmetry is odd. Wrap all sites or none.

**4. `mergeInheritedAVPs` memory**
Every child app gets a full copy of ancestor entries in both indices. For the shipped dictionaries this is fine, but worth a quick sanity check on allocation size after `Load()` for the Gx/Gy/S6a set; if it's meaningful, a comment noting the trade-off would help future readers.

**5. Minor**
- `RegisterDecoder` returning `ErrCannotUnregisterBuiltin` for nil-on-builtin is the right call, keep it.
- `IPv6Type` position in `datatype.go`; agreed, leave it at the end.

Main ask is #1. Once that's resolved I'll do a final pass and merge.

kulcsartibor:
Pushed the changes.
#1 `RegisterDecoder` mirrors built-in writes into Decoder so introspection stays accurate.
#4 Maybe there is a better approach to merge the dictionary with little less memory footprint. Most of diameter apps have base and app specific, so only the base is duplicated.

kulcsartibor:
For the #2 
Maybe in v5 or later it would be nicer to change the lookup as little. The `interface{}` if from before g 1.18 without generics

```go
// diam/dict/util.go — current
func (p *Parser) FindAVPWithVendor(appid uint32, code interface{}, vendorID uint32) (*AVP, error) {
    switch codeVal := code.(type) {
    case string:
        // ...
    case uint32:
        // ...
    case int:
        // ...
    default:
        return nil, fmt.Errorf("Unsupported AVP code type %T(%#v)", codeVal, code)
    }
}
```
Problems:
1. A `float64` or `[]byte` in code compiles but panics at runtime.
2. Every call boxes the argument (small heap allocation) and does a type switch (branch mispredict on the hot path).
3. IDEs can't suggest the right type at the call site.
4. The `int` case exists only because old Go code defaulted integer literals to `int`; modern callers don't need it.


Implement it like 
```go
type AVPCode interface {
    uint32 | string
}

func FindAVPWithVendor[T AVPCode](p *Parser, appid uint32, code T, vendorID uint32) (*AVP, error) {
    switch c := any(code).(type) {
    case uint32:
        return p.findByCode(appid, c, vendorID)
    case string:
        return p.findByName(appid, c, vendorID)
    }
    // Unreachable: the constraint rules out everything else at compile time.
    return nil, fmt.Errorf("dict: unreachable")
}
```

fiorix:
#1 is clean, thanks. Verified locally with behavioral tests on top of your branch: `Decoder` has all 19 built-ins at init, direct writes to `Decoder[builtin]` are ignored by `Decode` (array wins), `RegisterDecoder` on a built-in mirrors both array and map, restore works, nil-on-builtin returns `ErrCannotUnregisterBuiltin` without breaking the built-in. Dispatch order unchanged.

One subtlety: direct writes to `Decoder[builtin]` still mutate the map but are silently no-op'd by `Decode`. So `Decoder` can drift from real dispatch behavior if someone bypasses `RegisterDecoder`. The doc warns about this, which is enough.

#2 Agreed, v5 material. Generic constraint `uint32 | string` is the right shape; drop the `int` case in the process. Open an issue tagged v5 so it doesn't get lost?

#3 Still worth doing in this PR; 2-line change. Either wrap all three sites or drop `%w` from the two that have it. Consistency matters more than which direction you pick.

#4 Fine to ship as-is. A one-line comment on `mergeInheritedAVPs` noting the trade-off (base AVPs duplicated per child app, bounded by dictionary size at Load time, not per-message) would help future readers. The shipped dictionaries are small enough that duplication is noise.

Once #3 is fixed, please squash the eight commits into a single commit with a clean message (subject under 72 chars, body describing the perf wins + `RegisterDecoder` API + `FindAVPByCode`). I'll merge on green CI.

kulcsartibor:
Fixed and pushed the changes in squashed commit.

fiorix:
Squash and #3 look good. Two small things before I merge:

1. Please rebase onto current `main` (3 docs/cleanup commits landed today, none touch the files in this PR: `fd700a8`, `101c2e2`, `de48ccf`).
2. Move the "bounded duplication" note from the commit message into the `mergeInheritedAVPs` doc comment, e.g. one line like: `// Memory: base AVPs are duplicated per child app at Load time, not per message; bounded by dictionary size.` The commit-message version is fine but a future reader of the source won't see it.

Once those are in, I'll merge on green CI.

kulcsartibor:
Updated and squashed the commit. There is an issue with one test: `server_close_test.go:24`
It should run only in linux if sctp is available. I have a fix for but did not want to mix with the PR.

fiorix:
Rebase and #4 are clean. Verified locally: single commit on top of current main, doc comment now carries the memory note, all tests pass.

The `server_close_test.go` issue is pre-existing (came in via #231); please open a separate PR for that fix so we can keep this one focused.

Approving and will merge on green CI.

fiorix (approved):
LGTM. Will merge once CI is green.

PR #235, opened 2026-04-30
diam: prevent panic in DecodeGrouped on malformed AVP

IbrahimShahzad:
DecodeGrouped can panic if decoding the AVP fails and returns AVP{} and error. we next try to move ahead by avp.len() which would try dereferencing avp.Data which would be nil.


fiorix:
Thanks for catching this — the bug is real. `DecodeAVP` (diam/avp.go:55-61) always returns a non-nil `*AVP` even on error, so when decoding fails before `a.Data` is set, the next `avp.Len()` call derefs nil `Data` and panics. Fix is in the right place.

A few things to address before merge:

1. **Needs a regression test.** A panic fix without a test that reproduces the panic isn't really fixed — easy to add, e.g. call `DecodeGroupedFromBytes` with a payload too short for an 8-byte AVP header, or with a length field exceeding the buffer. The CI run shows no checks were triggered on the PR; a test in `diam/group_test.go` would also exercise CI.

2. **PR title.** "edge case check" is uninformative — and we're moving to squash-merges, so the PR title becomes the commit subject on main. Something like `diam: prevent panic in DecodeGrouped on malformed AVP` reads much better in `git log`.

3. **`avp == nil` is dead code.** `DecodeAVP` always allocates the pointer (avp.go:56) — only `avp.Data == nil` can fire. Drop the nil check or document why it's defensive.

4. **Comment is too long for a 1-line guard.** Something like `// avp.Len() would deref nil Data on partial decode` is plenty.

Optional follow-up (not for this PR): the root cause is `DecodeAVP` returning `(*AVP{}, err)` instead of `(nil, err)` on failure — that's the anti-pattern enabling this class of bug. Worth a separate PR since it's an API contract change.

IbrahimShahzad:
thanks for the detailed comments. ill update this.

fiorix:
Thanks for the updates — title and production fix look good. Two issues with the new test:

1. **Wrong function under test.** It calls `DecodeAVP`, but the panic guard you added is in `DecodeGroupedFromBytes`. The test should call `DecodeGroupedFromBytes(testGroupedAVPWithInvalidLen, 0, dict.Default)` to actually exercise the fix.

2. **Inverted assertion.** `if err != nil { t.Fatalf("expected error... got nil") }` fails on the success path. Should be `if err == nil`.

Also, before the fix, the test should panic (not just return an error) — worth confirming locally by reverting the guard and running the test to make sure it catches the regression.

CI hasn't run on this PR yet — I'll approve the workflow once the test is updated.

tridentsx:
I get this error not only on mallformed AVP but grouped AVP's without Dictionary, If I remember corectly the diameter standards are specific that AVP's without M-Bit set can contain anything and should be ignored by receiver if not understood. I will not ask AI to provide the references for this since last time it was a disaster.

But if I remember correctly

Unknown AVP with M-bit clear -> Ignore it in receiver
Unknown AVP with M-bit set -> Return an error, normally DIAMETER_AVP_UNSUPPORTED / 5001, with Failed-AVP
Unknown Grouped AVP with M-bit clear -> Can ignore the whole grouped AVP, even if it contains inner AVPs with M-bit set

fiorix:
Thanks, this is a useful distinction. I agree the M-bit behavior is a broader protocol-handling issue than this PR’s narrow panic guard.

For this PR, the immediate bug is that DecodeGrouped can panic when DecodeAVP returns a non-nil AVP with nil Data. The broader behavior around unknown AVPs should probably be handled separately: M-bit clear can be ignored/preserved, M-bit set should produce DIAMETER_AVP_UNSUPPORTED with Failed-AVP, and unknown non-mandatory grouped AVPs should not force decoding of their inner AVPs.

There is now a broader PR, #249, that overlaps this panic/malformed decode area. I’m going to review that first and then decide whether #235 is superseded.

fiorix:
I reviewed #249 and approved it. It covers this PR’s immediate panic class more broadly by preserving malformed known AVPs as `datatype.Unknown` so parse offsets stay aligned, and by guarding grouped decode fatal errors before `avp.Len()` can dereference nil `Data`.

If #249 lands, I think this PR is superseded. The M-bit / DIAMETER_AVP_UNSUPPORTED / Failed-AVP behavior from the discussion here should remain a separate protocol-handling follow-up rather than being folded into this narrow panic fix.

fiorix:
Thanks @IbrahimShahzad — this was the first report of the nil-pointer panic in `DecodeGrouped`/`DecodeGroupedFromBytes`, and your reproduction pinned it exactly (advancing by `avp.Len()` when the failed sub-AVP has `Data == nil`).

We consolidated the fix into #249, which addresses the same nil-deref **and** the related offset-desync from dictionary/wire type mismatches (e.g. an `Unsigned32` decoder silently zero-filling a wrong-length payload, whose `Len()` then misframes the parent loop and can trigger the very same panic downstream), with a broader regression suite. #249 is now merged.

Closing as superseded — but credit for surfacing this first. 🙏

IbrahimShahzad:
@fiorix Thanks! I'm really glad this got fixed. I also wanted to say I really appreciated how welcoming and constructive you have been. 
I'll definitely keep an eye out for anything else I can contribute in the future. Great work!

PR #240, opened 2026-04-30
sm: make Inband-Security-Id configurable in Client CER

### LINKED ISSUE #239: Client CER hardcodes Inband-Security-Id=0 — cannot propose TLS to peers (RFC 6733 §5.3.1)
## Problem

The sm.Client's `makeCER()` in `diam/sm/client.go` always includes `Inband-Security-Id=0`:

```go
m.NewAVP(avp.InbandSecurityID, avp.Mbit, 0, datatype.Unsigned32(0))
```

There is no way to configure the client to propose TLS (value 1) during CER/CEA exchange, even when the client intends to upgrade to TLS after the capabilities exchange.

## RFC Reference

**RFC 6733 §5.3.1** defines the CER message with:
> Inband-Security-Id: This field is used to indicate the security mechanisms supported by the local peer. The value 1 indicates TLS.

A client that supports in-band TLS SHOULD be able to include Inband-Security-Id=1 in its CER to propose TLS upgrade.

**RFC 6733 §2.1** (Transport):
> Diameter clients SHOULD support TLS/TCP and DTLS/SCTP.

## Suggested Fix

Add an `InbandSecurityID` field to the `sm.Client` struct:

```go
type Client struct {
    // ...existing fields...
    InbandSecurityID uint32 // 0 = NO_INBAND_SECURITY (default), 1 = TLS
}
```

Then in `makeCER()`:

```go
m.NewAVP(avp.InbandSecurityID, avp.Mbit, 0, datatype.Unsigned32(cli.InbandSecurityID))
```

This is backwards-compatible since the zero value preserves current behavior.

## Related

This issue is related to #236 (server-side rejection) and #238 (no post-CER TLS upgrade). All three would need to be addressed together for full RFC 6733 in-band TLS compliance.

tridentsx:
Add `InbandSecurityID` field to `sm.Client` allowing callers to propose TLS (value 1) during CER/CEA exchange per RFC 6733 §5.3.1. The zero value preserves existing behavior (NO_INBAND_SECURITY).

## Changes
- Add `InbandSecurityID uint32` field to `sm.Client` struct
- Use the field value in `makeCER()` instead of hardcoded `0`
- Add two tests verifying default (0) and TLS (1) values are sent

## RFC Reference
RFC 6733 §5.3.1 defines Inband-Security-Id AVP (code 299) with values 0 (NO_INBAND_SECURITY) and 1 (TLS).

All existing tests pass.

Fixes #239

### DIFF
--- a/diam/sm/client.go
+++ b/diam/sm/client.go
@@ -57,6 +57,7 @@ type Client struct {
 	AcctApplicationID           []*diam.AVP   // Acct applications
 	AuthApplicationID           []*diam.AVP   // Auth applications
 	VendorSpecificApplicationID []*diam.AVP   // Vendor specific applications
+	InbandSecurityID            uint32        // Inband-Security-Id for CER: 0=NO_INBAND_SECURITY (default), 1=TLS (RFC 6733 §5.3.1)
 }
 
 // Dial calls the address set as ip:port, performs a handshake and optionally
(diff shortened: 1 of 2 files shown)


fiorix:
LGTM. Minimal, backwards-compatible (zero-value preserves prior behavior), tests cover both the default and TLS=1 paths. Will merge.

PR #241, opened 2026-04-30
datatype: add DiameterURI Parse() method (RFC 6733 §4.3)

### LINKED ISSUE #237: DiameterURI datatype is opaque — no parsing of aaa:// vs aaas:// scheme (RFC 6733 §4.3)
## Problem

The `DiameterURI` datatype in `diam/datatype/diamuri.go` is implemented as a plain `OctetString` alias with no parsing:

```go
type DiameterURI OctetString

func DecodeDiameterURI(b []byte) (Type, error) {
    d := make([]byte, len(b))
    copy(d, b)
    return DiameterURI(OctetString(d)), nil
}
```

This means there is no way to programmatically extract the scheme (`aaa://` vs `aaas://`), FQDN, port, or transport parameters from a DiameterURI AVP value.

## RFC Reference

**RFC 6733 §4.3** defines the DiameterURI ABNF:

```
"aaa://" FQDN [ ":" port ] [ ";transport=" transport ] [ ";protocol=" protocol ]
"aaas://" FQDN [ ":" port ] [ ";transport=" transport ] [ ";protocol=" protocol ]
```

Key semantics:
- `aaa://` → default port 3868, unsecured transport
- `aaas://` → default port 5868, TLS-secured transport
- `;transport=tcp` or `;transport=sctp`

These fields carry meaning for peer discovery and redirect handling (e.g., Redirect-Host AVP).

## Impact

Applications using go-diameter cannot:
- Determine whether a Redirect-Host requires TLS by inspecting the scheme
- Extract the port or transport from a DiameterURI for connection establishment
- Validate DiameterURI format per the RFC ABNF

## Suggested Fix

Add a parsed representation:

```go
type ParsedDiameterURI struct {
    Secure    bool   // aaas:// = true, aaa:// = false
    FQDN      string
    Port      uint16 // default 3868 or 5868
    Transport string // "tcp" or "sctp"
    Protocol  string // "diameter", "radius", or "tacacsplus"
}

func (s DiameterURI) Parse() (*ParsedDiameterURI, error) { ... }
```

This would be backwards-compatible — the raw OctetString representation stays for wire encoding.

tridentsx:
Add a `Parse()` method to `DiameterURI` that extracts scheme (`aaa://` vs `aaas://`), FQDN, port, transport, and protocol per the RFC 6733 §4.3 ABNF.

## Changes
- Add `ParsedDiameterURI` struct with Secure, FQDN, Port, Transport, Protocol fields
- Add `Parse()` method to `DiameterURI` type
- Defaults: `aaa://` → port 3868, `aaas://` → port 5868, transport=tcp, protocol=diameter
- Add 10 test cases (valid URIs + error cases)

## RFC Reference
RFC 6733 §4.3 defines:
```
"aaa://" FQDN [ ":" port ] [ ";transport=" transport ] [ ";protocol=" protocol ]
"aaas://" FQDN [ ":" port ] [ ";transport=" transport ] [ ";protocol=" protocol ]
```

Backwards-compatible: the raw OctetString wire encoding is unchanged.

Fixes #237

### DIFF
--- a/diam/datatype/diamuri.go
+++ b/diam/datatype/diamuri.go
@@ -4,7 +4,11 @@
 
 package datatype
 
-import "fmt"
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
 
 // DiameterURI data type.
 type DiameterURI OctetString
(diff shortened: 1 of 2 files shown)


fiorix:
Logic is right, a few things to address before merge:

1. `indexOf` and `lastIndexByte` reimplement `strings.Index` / `strings.LastIndexByte`. Just import `strings` and drop the locals.
2. RFC 6733 §4.3: "Diameter URIs are case insensitive." The scheme prefix match (`raw[:7] == "aaas://"`) is case-sensitive. Use `strings.EqualFold` on the prefix, or lowercase before matching.
3. Port parsing should use `strconv.ParseUint(portStr, 10, 16)` — the current loop silently wraps on values > 65535.
4. Empty port (`host:`) currently parses as port 0; worth rejecting explicitly.

Happy either way: address here, or merge and follow up — your call.

PR #242, opened 2026-04-30
sm: accept Inband-Security-Id=1 when TLS is already established

### LINKED ISSUE #236: CER parser rejects Inband-Security-Id=1 (TLS) — violates RFC 6733 §5.3.1
## Problem

The CER parser in `diam/sm/smparser/cer.go` unconditionally rejects any CER containing `Inband-Security-Id=1` (TLS), returning `ErrNoCommonSecurity` (Result-Code 5017 NO_COMMON_SECURITY):

```go
if cer.InbandSecurityID != nil {
    if v := cer.InbandSecurityID.Data.(datatype.Unsigned32); v != 0 {
        return nil, ErrNoCommonSecurity
    }
}
```

This means peers that propose in-band TLS negotiation during CER/CEA exchange are always refused.

## RFC Reference

**RFC 6733 §5.3.1** defines the Inband-Security-Id AVP (code 299) with two values:
- 0: NO_INBAND_SECURITY
- 1: TLS

**RFC 6733 §6.2** states:
> If TLS is to be used with a Diameter connection, the TLS handshake MUST begin immediately following the CER/CEA exchange.

A compliant implementation should either:
1. Accept Inband-Security-Id=1 and initiate a TLS handshake after CEA, or
2. Respond with Inband-Security-Id=0 in CEA to indicate it only supports pre-established TLS (not reject the connection entirely)

## 3GPP Context

While 3GPP deployments (TS 33.210) typically use pre-established TLS/IPsec making this less critical in practice, RFC compliance requires at minimum not rejecting the connection when a peer proposes TLS.

## Suggested Fix

Instead of returning an error, the server should:
- If TLS is already established on the connection: accept the CER (security already satisfied)
- If TLS is not established and the server supports in-band TLS: negotiate TLS after CEA
- If TLS is not established and the server does NOT support in-band TLS: include Inband-Security-Id=0 in CEA (indicating no common security) but still allow the connection if the peer accepts

tridentsx:
Per RFC 6733 §5.3.1, a peer may include Inband-Security-Id=1 in CER to declare TLS capability. When the connection is already TLS-secured, this should be accepted rather than rejected with NO_COMMON_SECURITY.

## Changes
- Add `ParseWithSecurity(msg, role, tlsActive)` to CER parser
- Existing `Parse()` preserves backwards-compatible behavior (rejects on plain TCP)
- Update `handleCER` to pass `c.TLS() != nil` to the parser
- Add two unit tests

## RFC Reference
RFC 6733 §5.3.1, §6.2: When TLS is already established on the transport, Inband-Security-Id=1 simply declares the peer's TLS capability — the security requirement is already satisfied.

All existing tests pass (including TestHandleCER_InbandSecurity which verifies rejection on plain TCP).

Fixes #236

### DIFF
--- a/diam/sm/cer.go
+++ b/diam/sm/cer.go
@@ -28,7 +28,7 @@ func handleCER(sm *StateMachine) diam.HandlerFunc {
 			return
 		}
 		cer := new(smparser.CER)
-		_, err := cer.Parse(m, smparser.Server)
+		_, err := cer.ParseWithSecurity(m, smparser.Server, c.TLS() != nil)
 		if err != nil {
 			err = errorCEA(sm, c, m, cer, err)
 			if err != nil {
--- a/diam/sm/smparser/cer.go
+++ b/diam/sm/smparser/cer.go
@@ -31,6 +31,14 @@ type CER struct {
 // we don't support in our dictionary) and an error. Another cause
 // for error is the presence of Inband Security, we don't support that.
 func (cer *CER) Parse(m *diam.Message, localRole Role) (failedAVP *diam.AVP, err error) {
+	return cer.ParseWithSecurity(m, localRole, false)
+}
+
+// ParseWithSecurity is like Parse but accepts a tlsActive flag. When
+// tlsActive is true, Inband-Security-Id=1 (TLS) is accepted because
+// the transport is already secured — per RFC 6733 §5.3.1 the peer is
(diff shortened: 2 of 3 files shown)


fiorix:
LGTM. The `ParseWithSecurity` overload keeps `Parse` backwards-compatible and the existing `TestHandleCER_InbandSecurity` (rejection on plain TCP) still passes. Will merge.

PR #243, opened 2026-04-30
sm: implement post-CER/CEA TLS upgrade (RFC 6733 §6.2)

### LINKED ISSUE #238: No support for post-CER/CEA TLS upgrade (RFC 6733 §6.2)
## Problem

The server's connection handler in `diam/server.go` checks for TLS state only once at connection establishment:

```go
func (c *conn) serve() {
    // ...
    if tlsConn, ok := c.rwc.(*tls.Conn); ok {
        if err := tlsConn.Handshake(); err != nil {
            return
        }
        c.tlsState = &tls.ConnectionState{}
        *c.tlsState = tlsConn.ConnectionState()
    }
    for {
        m, err := c.readMessage()
        // ...
    }
}
```

There is no mechanism to upgrade a plain TCP connection to TLS after the CER/CEA exchange. Combined with the Inband-Security-Id=1 rejection (#236), this means in-band TLS negotiation as defined by RFC 6733 is completely unsupported.

## RFC Reference

**RFC 6733 §6.2** (Diameter Security):
> If TLS is to be used with a Diameter connection, the TLS handshake MUST begin immediately following the CER/CEA exchange for connections that are not pre-configured to use TLS.

The expected flow for in-band TLS:
1. TCP connection established (plain)
2. CER sent with Inband-Security-Id=1
3. CEA received with Inband-Security-Id=1
4. TLS handshake begins on the existing TCP connection
5. Diameter messages continue over TLS

## 3GPP Context

3GPP TS 33.210 (NDS/IP) mandates pre-established TLS or IPsec for inter-network-element communication, making in-band TLS upgrade less critical for telecom deployments. However, RFC compliance requires this capability for interoperability with non-3GPP Diameter peers.

## Suggested Approach

After CER/CEA exchange where both sides agree on Inband-Security-Id=1:
1. Pause message reading
2. Wrap `c.rwc` with `tls.Server(c.rwc, config)` (server side) or `tls.Client(c.rwc, config)` (client side)
3. Perform TLS handshake
4. Update `c.tlsState`
5. Resume message reading on the now-encrypted connection

This is similar to STARTTLS in SMTP/LDAP.

tridentsx:
Add support for upgrading a plain TCP connection to TLS after the CER/CEA exchange, as specified in RFC 6733 §6.2.

## Changes
- Add `TLSUpgrader` interface to `diam` package
- Add `StartTLS(*tls.Config) error` method to the server connection
- Add `TLSConfig *tls.Config` field to `sm.Settings`
- Update `handleCER`: when peer sends Inband-Security-Id=1 and server has TLSConfig, accept the CER, send success CEA, then upgrade to TLS
- Add `ParseWithSecurity()` and `RequestedSecurity()` to CER parser
- Add test verifying server accepts Inband-Security-Id=1 when TLSConfig is set

## RFC Reference
RFC 6733 §6.2:
> If TLS is to be used with a Diameter connection, the TLS handshake MUST begin immediately following the CER/CEA exchange for connections that are not pre-configured to use TLS.

## Backwards Compatibility
Without `TLSConfig` set in Settings, existing behavior is preserved (Inband-Security-Id=1 is rejected with NO_COMMON_SECURITY on plain connections).

All existing tests pass. The SCTP close test failure is pre-existing on upstream main.

Fixes #238
Related: #236, #239

### DIFF
--- a/diam/server.go
+++ b/diam/server.go
@@ -67,10 +67,15 @@ type CloseNotifier interface {
 //	    err := u.StartTLS(tlsConfig)
 //	}
 type TLSUpgrader interface {
-	// StartTLS upgrades the underlying connection to TLS. On the server
-	// side, cfg is used as tls.Server config; on the client side as
-	// tls.Client config. Must be called before any further reads/writes.
+	// StartTLS upgrades the underlying connection to TLS using
+	// server-side parameters (tls.Server). Must be called from within
+	// a handler before the next read.
 	StartTLS(cfg *tls.Config) error
+
+	// StartTLSClient upgrades the underlying connection to TLS using
+	// client-side parameters (tls.Client). Must be called from within
+	// a handler before the next read.
+	StartTLSClient(cfg *tls.Config) error
 }
 
 // A liveSwitchReader is a switchReader that's safe for concurrent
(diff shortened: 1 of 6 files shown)


fiorix:
Useful direction but a few things to resolve before merging:

1. **Server-only upgrade.** `StartTLS` lands on `*response` but there's no client-side counterpart. Without one, no in-tree peer can complete the upgrade — which is why the new test acknowledges it can't actually drive the handshake. RFC 6733 §6.2 requires both sides to upgrade.

2. **Concurrency.** `StartTLS` swaps `w.conn.rwc`, `w.conn.sr.r`, and `w.conn.buf`. The serve loop reads via `w.conn.buf` without holding `w.mu`. As long as the upgrade happens inside `handleCER` before the loop reads the next message it's probably safe in practice, but the invariant should be documented and asserted (e.g. only callable from a handler, before the next read).

3. **Failed upgrade after success CEA.** On `StartTLS` error the conn is closed, but the peer already received a success CEA. Either send a failure CEA on upgrade error, or document that an immediate close after success CEA means upgrade failure. (Doing the upgrade *before* the CEA contradicts §6.2.)

4. **Silent downgrade for non-upgradeable conns.** `canTLS := tlsActive || sm.cfg.TLSConfig != nil` accepts Inband-Security-Id=1 even when `c.(diam.TLSUpgrader)` fails the type assertion (e.g. SCTP, third-party transports). The upgrade is silently skipped and the conn stays plaintext after success CEA. Should reject the CER in that case.

Suggestion: split into (a) `TLSUpgrader` interface + server-side `StartTLS` as infra (this PR, scoped down), (b) client-side `StartTLS`, (c) `handleCER` orchestration with the safeguards above. Or hold this PR until (b) lands so we can ship them together with an end-to-end test.

#240, #241, #242 can land independently of this — they're useful on their own and don't block the design discussion here.

fiorix:
Thanks — the new commit addresses three of the four concerns nicely:

- ✅ **Client side**: `StartTLSClient` + `Client.RequestInbandTLS` + `handleCEA` upgrade closes the loop. Both peers can now complete the handshake.
- ✅ **Concurrency invariant**: documented on the interface ("must be called from within a handler before the next read"). Good.
- ✅ **No more silent downgrade**: `canTLS` now requires `c.(diam.TLSUpgrader)` to succeed, and the upgrade path also rejects defensively.

Two things left:

1. **#3 from the original review still open**: on `StartTLS` failure the conn is closed *after* a success CEA was sent — peer sees success-then-RST. Consider sending a failure CEA instead, or at minimum document that an immediate close after success CEA means upgrade failure.

2. **Rebase against main**: this PR is currently `mergeable=dirty` (#242 landed and conflicts with the parser changes here). Once rebased, CI will rerun and we can validate end-to-end with the new client-side path.

Also worth adding now that both sides are wired up: an integration test that drives a full plain→CER/CEA→TLS handshake using the new `StartTLSClient`. That would have been impossible before this revision and is the strongest proof the feature works.

tridentsx:
I tried to address those last things please check if taths ok !

PR #246, opened 2026-05-09
sm: include Inband-Security-Id in CEA when TLS upgrade accepted

### LINKED ISSUE #245: sm: CEA missing Inband-Security-Id AVP during TLS upgrade negotiation
## Problem

When a peer sends CER with `Inband-Security-Id=1` requesting TLS upgrade, the server sends a success CEA but does **not** include `Inband-Security-Id=1` in the response. Per RFC 3588 §6.2 (and RFC 6733 backward compatibility), the answering peer must echo the negotiated Inband-Security-Id in the CEA so the initiator knows TLS upgrade was accepted.

Without this AVP, compliant peers (e.g. Ericsson Diameter Load Balancer) interpret the CEA as rejecting TLS and disconnect immediately — before the server-side `StartTLS` handshake can begin.

## Observed Behavior

1. Peer connects plain TCP, sends CER with `Inband-Security-Id=1`
2. Server parses CER, calls `successCEA()` — CEA is sent **without** Inband-Security-Id
3. Server calls `StartTLS()` → `tls.Server().Handshake()` → returns EOF
4. Peer disconnected because it didn't see Inband-Security-Id=1 in CEA

## Expected Behavior

The CEA should include `Inband-Security-Id=1` when the server agrees to upgrade. Both peers then proceed with TLS handshake.

## Fix

In `successCEA()` (`diam/sm/cer.go`), add:
```go
if cer.RequestedSecurity() == 1 && sm.cfg.TLSConfig != nil {
    a.NewAVP(avp.InbandSecurityID, avp.Mbit, 0, datatype.Unsigned32(1))
}
```

## Related

This is a follow-up to PR #243 which implemented the TLS upgrade mechanics but missed the CEA AVP.

tridentsx:
When the server accepts `Inband-Security-Id=1` from a CER and has `TLSConfig` set, echo the AVP in the success CEA. Without this, the initiating peer cannot know TLS was accepted and disconnects before the TLS handshake begins.

Per RFC 3588 §6.2, the CEA must contain the negotiated `Inband-Security-Id` so both peers agree on whether to proceed with TLS.

Fixes #245

### DIFF
--- a/diam/sm/cer.go
+++ b/diam/sm/cer.go
@@ -140,6 +140,10 @@ func errorCEA(sm *StateMachine, c diam.Conn, m *diam.Message, cer *smparser.CER,
 	if sm.cfg.FirmwareRevision != 0 {
 		a.NewAVP(avp.FirmwareRevision, 0, 0, sm.cfg.FirmwareRevision)
 	}
+	// Echo Inband-Security-Id when TLS upgrade was negotiated (RFC 3588 §6.2).
+	if cer.RequestedSecurity() == 1 && sm.cfg.TLSConfig != nil {
+		a.NewAVP(avp.InbandSecurityID, avp.Mbit, 0, datatype.Unsigned32(1))
+	}
 	_, err = a.WriteTo(c)
 	if err != nil {
 		err = fmt.Errorf("Error CEA '%s' send failure: %v", errMessage, err)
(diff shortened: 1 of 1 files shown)


fiorix:
Closing — the post-CER/CEA TLS upgrade feature this builds on (#243) was reverted in #244 per @vingarzan's RFC 6733 §5.6 / §6.10 feedback on #238. The pre-existing TLS-on-listen path (TLS handshake before any Diameter messages) remains the spec-compliant approach. If a legitimate need for in-band TLS negotiation surfaces, please discuss on #238 first.

PR #252, opened 2026-06-24
sm: retain CEA capability AVPs (Vendor-Id, Product-Name, Supported-Vendor-Id) on smpeer.Metadata

gergogera:
## Summary

The `smpeer` package already persists peer handshake `Metadata` on the connection context — `handleCER`/`handleCEA` call `c.SetContext(smpeer.NewContext(...))` so callers can read peer info after the handshake. Today it keeps only `Origin-Host`, `Origin-Realm` and the application list, and the parsed `CER`/`CEA` is then discarded.

This PR extends that existing mechanism so the remaining peer **capability AVPs** are reachable post-handshake:

- **`smparser.CEA`** gains:
  - `Vendor-Id` (`uint32`) — **mandatory** in the CEA grammar (RFC 6733 §5.3.2)
  - `Product-Name` (`string`) — **mandatory** (RFC 6733 §5.3.2)
  - `Supported-Vendor-Id` (`[]*diam.AVP`) — standard, optional, repeatable

  All parsed through the existing `avp:`-tag unmarshalling — no new parsing logic.
- **`smpeer.Metadata`** gains `Cer *smparser.CER` / `Cea *smparser.CEA` pointers to the parsed handshake message, populated by `FromCER`/`FromCEA`, so a consumer holding the connection can read any capability AVP the peer advertised.

## Motivation

`smpeer`'s stated purpose is *"extract information from a CER or CEA, and associate with a Context."* Persisting peer metadata on the connection is already the package's accepted design — this change simply stops it from dropping the (RFC-mandatory) `Vendor-Id`/`Product-Name` and the standard `Supported-Vendor-Id`, which a capabilities-introspection consumer otherwise cannot recover once the handshake completes.

## Compatibility

Purely additive:
- Existing `Metadata` fields and all current behaviour are unchanged.
- The new `Cer`/`Cea` pointers are `nil` unless the corresponding constructor is used.
- No signature changes; no on-wire change.

## Tests

- `smparser`: `TestCEACapabilityAVPs` parses a CEA carrying `Vendor-Id`, `Product-Name` and two `Supported-Vendor-Id` AVPs and asserts the new fields.
- `smpeer`: `TestFromCER`/`TestFromCEA` extended to assert the correct pointer is populated and the other is `nil`.
- `gofmt`/`go vet` clean; `go test -race -cover ./diam/sm/...` green (`smparser` 89.7%, `smpeer` 100%).

### DIFF
--- a/diam/sm/smparser/cea.go
+++ b/diam/sm/smparser/cea.go
@@ -18,6 +18,9 @@ type CEA struct {
 	OriginHost                  datatype.DiameterIdentity `avp:"Origin-Host"`
 	OriginRealm                 datatype.DiameterIdentity `avp:"Origin-Realm"`
 	OriginStateID               uint32                    `avp:"Origin-State-Id"`
+	VendorID                    uint32                    `avp:"Vendor-Id"`           // Peer's Vendor-Id (RFC 6733 §5.3.2, mandatory).
+	ProductName                 string                    `avp:"Product-Name"`        // Peer's Product-Name (RFC 6733 §5.3.2, mandatory).
+	SupportedVendorID           []*diam.AVP               `avp:"Supported-Vendor-Id"` // Vendor-Ids the peer supports (RFC 6733 §5.3.2).
 	AcctApplicationID           []*diam.AVP               `avp:"Acct-Application-Id"`
 	AuthApplicationID           []*diam.AVP               `avp:"Auth-Application-Id"`
 	VendorSpecificApplicationID []*diam.AVP               `avp:"Vendor-Specific-Application-Id"`
--- a/diam/sm/smparser/cea_test.go
+++ b/diam/sm/smparser/cea_test.go
@@ -260,6 +260,45 @@ func TestCommonAppIdCEA(t *testing.T) {
 	}
 }
 
+// TestCEACapabilityAVPs verifies that the capability AVPs from RFC 6733 §5.3.2
+// (Vendor-Id, Product-Name, Supported-Vendor-Id) are parsed into the CEA struct.
+func TestCEACapabilityAVPs(t *testing.T) {
+	m := diam.NewMessage(diam.CapabilitiesExchange, 0, 0, 0, 0, nil)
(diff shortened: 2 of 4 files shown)


fiorix (changes requested):
Thanks for this. The implementation is small and the focused checks pass locally:

- `go test ./diam/sm/...`
- `go test -race ./diam/sm/...`
- `go vet ./diam/sm/...`

One small API-shape request before merge: could you rename the new exported `smpeer.Metadata` fields from `Cer`/`Cea` to `CER`/`CEA`? The rest of this package already uses the protocol initialisms in exported names (`FromCER`, `FromCEA`, `smparser.CER`, `smparser.CEA`), and this is new public API, so it is worth getting the names right before release.

Also, the comment currently says both retained pointers expose capability AVPs such as `Vendor-Id`, `Product-Name`, and `Supported-Vendor-Id`, but only `smparser.CEA` gains those fields in this PR. I would either narrow that comment to CEA capabilities or add matching CER fields if inbound peer capability access is intended as well.

gergogera:
Thanks for the review! Both addressed in 610bef0:

1. Renamed the new exported `smpeer.Metadata` fields `Cer`/`Cea` → `CER`/`CEA` (and the test assertions), matching `FromCER`/`FromCEA` and `smparser.CER`/`smparser.CEA`.
2. Narrowed the doc comment so only `CEA` is described as exposing the capability AVPs (`Vendor-Id`, `Product-Name`, `Supported-Vendor-Id`) added in this PR; `CER` is documented as retaining the parsed inbound message without claiming those fields.

`go vet ./diam/sm/...` clean and `go test -race -cover ./diam/sm/...` green (`smpeer` 100%).

fiorix (approved):
Thanks for the update. This addresses my review comments: the exported metadata fields now use the `CER`/`CEA` initialisms, and the comment now matches the CEA capability fields exposed by the PR.

I reran the focused checks locally on `610bef0`:

- `go test ./diam/sm/...`
- `go test -race ./diam/sm/...`
- `go vet ./diam/sm/...`

All passed. Approved.

PR #253, opened 2026-06-24
sm: add OnCEA/OnDWA server-side handshake hooks (follow-up to #233)

gergogera:
## Summary

PR #233 added `Settings.OnCER` and `Settings.OnDWR` pre-hooks for server-side **inbound** CER and DWR. This adds the symmetric **outbound** pair:

- **`OnCEA`** — invoked immediately before a CEA is sent, on the fully-constructed answer. Fires for both the success CEA (`successCEA`) and the error CEA (`errorCEA`).
- **`OnDWA`** — invoked immediately before a DWA is sent in response to a peer DWR (`handleDWR`).

Both are nil-guarded `diam.HandlerFunc` fields on `Settings`. Default behaviour is unchanged when nil — pure additive, no signature changes, no behaviour change for existing users.

## Why a call-site hook (not `chainPreHook`)

CEA and DWA are **built and sent by the state machine internally**, and `sm.HandleFunc`/`HandleIdx` explicitly refuse `CER`/`CEA`/`DWR` handler registration to protect the handshake. So the server's outbound handshake answers have no user interception point — exactly the situation #233 solved for inbound CER/DWR. `chainPreHook` only applies to *received* messages dispatched on the mux; for an answer the SM constructs and writes itself, a nil-guarded call-site insert before `a.WriteTo(c)` is the minimal mechanism.

## Changes

| File | Change |
|---|---|
| `diam/sm/sm.go` | `OnCEA`/`OnDWA` fields on `Settings` (paired with `OnCER`/`OnDWR`), doc comments in the same style |
| `diam/sm/cer.go` | nil-guarded `OnCEA` call before send in `successCEA` **and** `errorCEA` |
| `diam/sm/dwr.go` | nil-guarded `OnDWA` call before send in `handleDWR` |
| `diam/sm/hooks_test.go` | `TestOnCEAHook`, `TestOnCEAHookErrorCEA`, `TestOnDWAHook` — mirroring the `TestOnCERHook`/`TestOnDWRHook` template |

## Testing

- `go test -race ./diam/sm/` passes; the three new tests cover the success-CEA, error-CEA (forced via an unsupported `Acct-Application-Id` → `DIAMETER_NO_COMMON_APPLICATION`), and DWA paths.
- `go vet ./...` clean; `gofmt` clean.
- Default-behaviour-unchanged is covered by the existing handshake tests passing unmodified.

### DIFF
--- a/diam/sm/cer.go
+++ b/diam/sm/cer.go
@@ -104,6 +104,9 @@ func errorCEA(sm *StateMachine, c diam.Conn, m *diam.Message, cer *smparser.CER,
 	if sm.cfg.FirmwareRevision != 0 {
 		a.NewAVP(avp.FirmwareRevision, 0, 0, sm.cfg.FirmwareRevision)
 	}
+	if sm.cfg.OnCEA != nil {
+		sm.cfg.OnCEA(c, a)
+	}
 	_, err = a.WriteTo(c)
 	if err != nil {
 		err = fmt.Errorf("Error CEA '%s' send failure: %v", errMessage, err)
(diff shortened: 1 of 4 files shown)


fiorix (approved):
Thanks, this looks well scoped. I reviewed the call sites for success CEA, error CEA, and DWA, and the focused checks pass locally:

- `go test ./diam/sm/...`
- `go test -race ./diam/sm/...`
- `go vet ./diam/sm/...`

No blocking comments from my pass.

PR #256, opened 2026-07-01
diam: fix nil-pointer panic decoding malformed grouped AVPs

gen2brain:
DecodeGroupedFromBytes and the top-level decodeAVPs loop advanced the read offset by avp.Len() even when DecodeAVP returned an error.
On an error, the AVP's Data can be nil, and AVP.Len() dereferences it, so a grouped AVP whose sub-AVP fails to decode panics instead of returning an error.

Both loops now stop on the first decode error, returning the partial result plus the error.

### DIFF
--- a/diam/group.go
+++ b/diam/group.go
@@ -35,7 +35,10 @@ func DecodeGroupedFromBytes(b []byte, application uint32, dictionary *dict.Parse
 	for n := 0; n < len(b); {
 		avp, err := DecodeAVP(b[n:], application, dictionary)
 		if err != nil {
+			// Stop: framing is now untrustworthy and avp.Data may be nil, so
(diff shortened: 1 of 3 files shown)


fiorix:
Thanks @gen2brain — this is a correct fix for the nil-panic, and you covered both loops (`DecodeGroupedFromBytes` and `Message.decodeAVPs`), which #235 didn't.

We went with #249 (now merged) to consolidate, since it also handles the offset-desync from dictionary/wire type mismatches — the root cause that can misframe the stream and trigger the same panic further along — while keeping `DecodeError` sub-AVPs recoverable rather than breaking on every error. Verified against the same malformed-grouped-AVP input that panicked on main; it now returns an error cleanly.

Closing as superseded by #249.

PR #257, opened 2026-07-09
diam: don't close the connection on an unregistered application/command

towi:
## Summary

- `readHeader`'s `FindCommand` lookup hard-fails for any `(application, command)` pair the loaded dictionary doesn't know about — e.g. an application whose XML dictionary isn't part of `dict.Default` at all, such as 3GPP Cx/Dx (16777216). `ReadMessage` then returns an error for an otherwise well-formed message.
- `conn.serve()` responds to *any* `ReadMessage` error by closing the whole connection (`c.rwc.Close(); break`) — so one message on an unregistered application takes down the entire peer link, including every other request in flight on it.
- Fix: fall back to a permissive `Command` instead of erroring, so `decodeAVPs` still runs. AVP-level lookups already degrade to `datatype.Unknown` for codes the dictionary doesn't recognize (`dict.FindAVPByCode`), so the message decodes without error — its AVPs just come back untyped when the application itself is unregistered.

Found this while load-testing a Diameter HSS: a client advertising both S6a/SWx (which `dict.Default` ships) and Cx/Dx (which it doesn't) in its CER would have its whole connection killed by a single Cx answer, even though the peer sent a perfectly valid message.

## Test plan

- [x] Added `TestReadMessageUnregisteredApplication`, reproducing the crash with a well-formed message on an unregistered application-id, asserting `ReadMessage` no longer errors.
- [x] `go test ./diam/...` passes (existing suite untouched).
- [x] `go vet ./diam/...` clean.
- [x] Verified against a real Diameter peer via [xk6-diameter](https://github.com/lwlee2608/xk6-diameter) load-testing a Cx MAR/MAA exchange against a live HSS — went from every request timing out (connection silently torn down) to 100% success.

### DIFF
--- a/diam/message.go
+++ b/diam/message.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log"
 	"math/rand"
 	"net"
 	"strings"
@@ -23,6 +24,19 @@ import (
 
 // MessageBufferLength is the default buffer length for Diameter messages.
 var MessageBufferLength = 1 << 10
+
+// unknownCommand is used when the dictionary has no application/command
+// entry for a received message (e.g. an application whose dictionary isn't
+// loaded at all). Its permissive rule set lets decodeAVPs run generically
+// instead of aborting -- AVP-level lookups already degrade to
+// datatype.Unknown for codes the dictionary doesn't recognize
+// (dict.FindAVPByCode), so a message like this still decodes instead of
+// erroring out, even though its AVPs come back untyped.
+var unknownCommand = &dict.Command{
+	Name:    "Unknown",
+	Request: dict.CommandRule{Rule: make([]*dict.Rule, 32)},
+	Answer:  dict.CommandRule{Rule: make([]*dict.Rule, 32)},
+}
 
 func init() {
 	rand.Seed(time.Now().UnixNano())
@@ -113,7 +127,9 @@ func (m *Message) readHeader(r io.Reader, buf *bytes.Buffer) (cmd *dict.Command,
 		m.Header.CommandCode,
 	)
 	if err != nil {
-		return nil, stream, err
+		log.Printf("diam: no dictionary command for app %d code %d, decoding generically: %v",
+			m.Header.ApplicationID, m.Header.CommandCode, err)
+		cmd = unknownCommand
 	}
 	return cmd, stream, nil
 }
--- a/diam/message_test.go
+++ b/diam/message_test.go
@@ -152,6 +152,59 @@ func TestReadMessageWithVendorID(t *testing.T) {
 	t.Logf("Message:\n%s", msg)
 }
 
+// TestReadMessageUnregisteredApplication covers a message whose
+// Application-Id has no dictionary entry at all (e.g. 3GPP Cx/Dx,
+// 16777216, isn't part of dict.Default). Before this fix, readHeader's
+// FindCommand lookup failed outright for ANY message on such an
+// application -- even a well-formed one carrying only ordinary AVPs -- and
+// ReadMessage returned an error. Since conn.serve() closes the whole
+// connection on any ReadMessage error, a single message on an unregistered
+// application used to take down the entire peer connection, not just that
+// one message.
+//
+// FindAVPByCode doesn't chain to the base application (id 0) for an
+// unregistered app, so even well-known AVPs like Result-Code decode as
+// datatype.Unknown here rather than their typed dictionary value -- decoding
+// no longer errors out, which is what this fix addresses, but recovering
+// typed values for base AVPs on unregistered applications is a separate
+// concern this change doesn't attempt to solve.
+func TestReadMessageUnregisteredApplication(t *testing.T) {
+	const unregisteredApp = 16777216 // 3GPP Cx/Dx, not in dict.Default
+	const someCommand = 303          // Multimedia-Auth-Request/Answer
+
+	m := NewMessage(someCommand, 0, unregisteredApp, 1, 1, dict.Default)
+	if _, err := m.NewAVP(avp.SessionID, avp.Mbit, 0, datatype.UTF8String("cx;1;1")); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := m.NewAVP(avp.ResultCode, avp.Mbit, 0, datatype.Unsigned32(2001)); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := m.NewAVP(avp.OriginHost, avp.Mbit, 0, datatype.DiameterIdentity("hss")); err != nil {
+		t.Fatal(err)
+	}
+
+	var buf bytes.Buffer
+	if _, err := m.WriteTo(&buf); err != nil {
+		t.Fatal(err)
+	}
+
+	got, err := ReadMessage(&buf, dict.Default)
+	if err != nil {
+		t.Fatalf("ReadMessage should decode generically for an unregistered application, got error: %v", err)
+	}
+	if len(got.AVP) != 3 {
+		t.Fatalf("got %d AVPs, want 3", len(got.AVP))
+	}
+	rc, err := got.FindAVP(avp.ResultCode, 0)
+	if err != nil {
+		t.Fatalf("Result-Code AVP not found: %v", err)
+	}
+	want := datatype.Unknown{0x00, 0x00, 0x07, 0xd1} // 2001, big-endian
+	if got, ok := rc.Data.(datatype.Unknown); !ok || !bytes.Equal(got, want) {
+		t.Fatalf("Result-Code raw bytes = %#v, want %#v", rc.Data, want)
+	}
+}
+
 func TestNewMessage(t *testing.T) {
 	want, _ := ReadMessage(bytes.NewReader(testMessage), dict.Default)
 	m := NewMessage(CapabilitiesExchange, RequestFlag, 0, 0xa8cc407d, 0xa8c1b2b4, dict.Default)


towi:
## Minimal repro

Self-contained, no sipgate-specific anything: a stock go-diameter server + a k6/xk6-diameter client in one image. Confirmed on unpatched `main` — 2 of 3 checks fail, one message on an unregistered application (16777216, 3GPP Cx/Dx) kills the whole connection. With this PR's fix applied to both sides, all 3 pass.

**`Dockerfile`**
```dockerfile
COMMENT_DOCKERFILE_EOF
# Minimal, self-contained repro for the go-diameter connection-close bug.
# Builds a stock (unpatched) go-diameter server and a k6+xk6-diameter
# client into one image; the entrypoint runs both and prints the result.
#
# Build & run:
#   docker build -t godiam-repro .
#   docker run --rm godiam-repro

FROM golang:1.25 AS server-build
WORKDIR /src
RUN go mod init repro && go get github.com/fiorix/go-diameter/v4@main
COPY server.go .
RUN go mod tidy && go build -o /out/server server.go

FROM grafana/xk6 AS client-build
RUN git clone https://github.com/lwlee2608/xk6-diameter.git /xk6/xk6-diameter && \
    git -C /xk6/xk6-diameter checkout 48db2efd4febbcb756a06218f74c5101f7eb37ed
RUN xk6 build latest \
    --with github.com/matrixxsoftware/xk6-diameter=/xk6/xk6-diameter \
    --output /xk6/k6

FROM debian:trixie-slim
RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates && \
    rm -rf /var/lib/apt/lists/*
COPY --from=server-build /out/server /usr/bin/diam-server
COPY --from=client-build /xk6/k6 /usr/bin/k6
COPY repro.js /repro.js
COPY entrypoint.sh /entrypoint.sh
RUN chmod +x /entrypoint.sh
ENTRYPOINT ["/entrypoint.sh"]
```

**`server.go`**
```go
// Minimal Diameter server, stock go-diameter, no custom dictionary. Answers
// any request it manages to decode with a generic Success. Demonstrates
// that receiving a single well-formed message on an application-id the
// dictionary doesn't know about (e.g. 16777216, 3GPP Cx/Dx) kills the whole
// connection -- not just that one request.
package main

import (
	"log"

	"github.com/fiorix/go-diameter/v4/diam"
	"github.com/fiorix/go-diameter/v4/diam/avp"
	"github.com/fiorix/go-diameter/v4/diam/datatype"
	"github.com/fiorix/go-diameter/v4/diam/sm"
)

func main() {
	settings := &sm.Settings{
		OriginHost:       datatype.DiameterIdentity("server"),
		OriginRealm:      datatype.DiameterIdentity("go-diameter"),
		VendorID:         13,
		ProductName:      "go-diameter-repro-server",
		FirmwareRevision: 1,
	}
	mux := sm.New(settings)
	mux.HandleFunc("ALL", func(c diam.Conn, m *diam.Message) {
		log.Printf("received:\n%s", m)
		a := m.Answer(diam.Success)
		a.NewAVP(avp.OriginHost, avp.Mbit, 0, settings.OriginHost)
		a.NewAVP(avp.OriginRealm, avp.Mbit, 0, settings.OriginRealm)
		if _, err := a.WriteTo(c); err != nil {
			log.Printf("write answer: %v", err)
			return
		}
		log.Printf("sent:\n%s", a)
	})
	go func() {
		for report := range mux.ErrorReports() {
			log.Printf("mux error report: %v", report)
		}
	}()
	log.Println("listening on :3868")
	log.Fatal(diam.ListenAndServe(":3868", mux, nil))
}
```

**`entrypoint.sh`**
```sh
#!/bin/sh
set -e
/usr/bin/diam-server &
SERVER_PID=$!
sleep 1
k6 run /repro.js
STATUS=$?
kill $SERVER_PID 2>/dev/null || true
exit $STATUS
```

**`repro.js`**
```js
// Reproduces: a single well-formed message on an application-id the
// server's dictionary doesn't know about (e.g. 16777216, 3GPP Cx/Dx --
// not part of dict.Default) kills the WHOLE connection, not just that
// one request. Uses xk6-diameter (github.com/lwlee2608/xk6-diameter,
// built on fiorix/go-diameter) purely as a Diameter client; the bug
// itself lives in go-diameter, triggered here against a stock,
// unpatched go-diameter server (server.go, same repro bundle).
//
// Expected on unpatched go-diameter: step 1 succeeds, step 2 hangs/errors
// (the server's conn.serve() already died decoding the request header --
// FindCommand fails for the unregistered app-id), step 3 fails too, on
// the SAME connection that step 1 used successfully.
import diam from "k6/x/diameter";
import avpmod from "k6/x/diameter/avp";
import { check } from "k6";

export const options = {
  scenarios: {
    smoke: { executor: "per-vu-iterations", vus: 1, iterations: 1, maxDuration: "30s" },
  },
};

const TARGET = __ENV.TARGET || "127.0.0.1:3868";
const data = diam.DataType();

const CMD_R = 0x80; // Request bit
const FLAG_M = 0x40; // Mandatory
const APP_CREDIT_CONTROL = 4; // Charging Control -- common app both sides advertise
const APP_UNREGISTERED = 16777216; // 3GPP Cx/Dx -- not in dict.Default
const CMD_CCR = 272; // Credit-Control-Request, well-known under APP_CREDIT_CONTROL
const AVP_SESSION_ID = 263;
const AVP_ORIGIN_HOST = 264;
const AVP_RESULT_CODE = 268;
const AVP_ORIGIN_REALM = 296;
const AVP_DESTINATION_REALM = 283;

function newClient() {
  return diam.Client({
    requestTimeout: "5s",
    enableWatchdog: false,
    authApplicationId: [APP_CREDIT_CONTROL],
    capabilityExchange: {
      vendorId: 13,
      productName: "xk6-diameter-repro",
      originHost: "client.repro",
      originRealm: "repro.local",
    },
  });
}

// A minimal, well-formed request on the given (cmdCode, appId).
function request(cmdCode, appId, sessionSuffix) {
  const m = diam.newMessage(cmdCode, appId, CMD_R);
  m.add(avpmod.New(AVP_SESSION_ID, 0, FLAG_M, data.UTF8String("repro;1;" + sessionSuffix)));
  m.add(avpmod.New(AVP_ORIGIN_HOST, 0, FLAG_M, data.DiameterIdentity("client.repro")));
  m.add(avpmod.New(AVP_ORIGIN_REALM, 0, FLAG_M, data.DiameterIdentity("repro.local")));
  m.add(avpmod.New(AVP_DESTINATION_REALM, 0, FLAG_M, data.DiameterIdentity("repro.local")));
  return m;
}

// Sends msg and reports whether an answer with Result-Code 2001 came back.
function trySend(client, msg, label) {
  try {
    const answer = client.send(msg);
    const rc = answer.findAVP(AVP_RESULT_CODE, 0);
    console.log(`${label}: OK, result-code=${rc}`);
    return Number(rc) === 2001;
  } catch (e) {
    console.log(`${label}: FAILED -- ${e}`);
    return false;
  }
}

// Like trySend, but only asks "did an answer come back at all" -- used for
// the unregistered-app request, whose Result-Code decodes as untyped
// datatype.Unknown rather than 2001 (FindAVPByCode doesn't chain to the
// base app for an app-id it doesn't know), so a typed comparison doesn't
// apply here. Not throwing is exactly what this fix guarantees.
function trySendExpectNoThrow(client, msg, label) {
  try {
    const answer = client.send(msg);
    console.log(`${label}: OK, got an answer (${answer.findAVP(AVP_RESULT_CODE, 0)})`);
    return true;
  } catch (e) {
    console.log(`${label}: FAILED -- ${e}`);
    return false;
  }
}

export default function () {
  const client = newClient();
  try {
    client.connect(TARGET);
  } catch (e) {
    console.log(`connect: FAILED -- ${e}`);
    check(null, { "connected": () => false });
    return;
  }

  const step1 = trySend(client, request(CMD_CCR, APP_CREDIT_CONTROL, 1), "1. baseline request (known app)");
  check(null, { "1. baseline request succeeds": () => step1 });

  const step2 = trySendExpectNoThrow(client, request(303, APP_UNREGISTERED, 2), "2. request on unregistered app-id");
  check(null, { "2. unregistered-app request gets an answer": () => step2 });

  // Same client/connection as step 1 -- proves the bad message in step 2
  // took the whole connection down, not just its own request.
  const step3 = trySend(client, request(CMD_CCR, APP_CREDIT_CONTROL, 3), "3. baseline request again, same connection");
  check(null, { "3. connection survives the bad message": () => step3 });
}
```

**Run:**
```
docker build -t godiam-repro .
docker run --rm godiam-repro
```

**Unpatched (`main`):**
```
✓ 1. baseline request succeeds
✗ 2. unregistered-app request gets an answer
✗ 3. connection survives the bad message
checks..................: 33.33% ✓ 1  ✗ 2
```

**With this fix (both server and client built against the fix branch):**
```
✓ 1. baseline request succeeds
✓ 2. unregistered-app request gets an answer
✓ 3. connection survives the bad message
checks..................: 100.00% ✓ 3  ✗ 0
```

fiorix:
One cleanup before this is ready: please remove the `log.Printf` in `readHeader`. Unknown commands can be normal peer traffic once this fallback exists, and decoding should not write to the package-global logger on the successful path. Surface it through existing handler/error/reporting paths only if the caller needs it.

I verified the branch with `go test ./diam/...`; it passed. The generic decode behavior looks narrowly scoped.

volkland:
(responding because towi is a colleague and we're working on the same topic)

> please remove the log.Printf in readHeader. Unknown commands can be normal peer traffic once this fallback exists, and decoding should not write to the package-global logger on the successful path. Surface it through existing handler/error/reporting paths only if the caller needs it.

I agree with removing the log line, but do not understand how you would surface the error without interrupting the control flow? Maybe only allow this fallback in a config value (or add option to explicitly not fallback/return an error)? I fear that this PR is a change in behavior and would break library-users that rely on failing when the diameter command is unknown.

fiorix:
Yes, this is a behavior change at the `ReadMessage` boundary: an unknown `(Application-Id, Command-Code)` now decodes generically instead of returning the dictionary lookup error.

I think that is the right behavior here. A command dictionary miss means "this local dictionary does not know the command", not "the wire message is malformed". Returning that as a read error is what makes `conn.serve()` close the whole peer connection, which is the bug this PR fixes.

On surfacing without interrupting control flow: that path already exists. `ServeMux` does its own `FindCommand` lookup after decode and routes unknown commands to the `ALL` handler if one is registered, otherwise it emits an `ErrorReport` on the `ErrorReports` channel, without killing the connection. So even with the `log.Printf` removed, an unknown command is not silent: with no `ALL` handler it lands in `ErrorReports` as an unhandled message, and towi's repro server already drains exactly that channel. Callers that need the old strict behavior can reproduce it after decode with `m.Dictionary().FindCommand(m.Header.ApplicationID, m.Header.CommandCode)`, the same check `ReadMessage` used to fail on, now in the caller's hands.

One caveat worth being explicit about: on an unregistered application the AVPs decode untyped (`datatype.Unknown`), since `FindAVPByCode` does not chain to the base application for an app id the dictionary doesn't know. The message survives decode, but recovering typed values for base AVPs there is a separate concern this PR intentionally doesn't take on. Towi's repro notes the same.

So I do not think this needs a config knob unless someone has a concrete compatibility case we need to preserve. The one thing I still want changed before merge is the `log.Printf` in `readHeader`; unknown commands can be normal traffic, and successful decoding should stay quiet.

volkland:
Thanks for clarifying, that all makes sense, and the ALL handler / ErrorReports is a good sink for those messages.

towi won't be back for about a month, if you're keen on merging soon, I'd suggest that you add a commit to the branch removing the logline.

Thanks for your work maintaining the package.

